PR #8, opened 2019-10-16
Replace EL with scriptlets in SkylineToolsStore jsp files

kaipot:


SkylineToolsStore/src/org/labkey/skylinetoolsstore/view/SkylineToolDetails.jsp
```diff
@@ -31,7 +31,7 @@
     final String jsDir = contextPath + "/skylinetoolsstore/js/";
 
     final String autocompleteUsers = admin ? SkylineToolsStoreController.getUsersForAutocomplete() : "\"\"";
-    pageContext.setAttribute("autocompleteUsers", autocompleteUsers);
+    // pageContext.setAttribute("autocompleteUsers", autocompleteUsers);
```
labkey-jeckels: Suggest deleting this instead of leaving it commented out.

PR #22, opened 2020-01-08
Explicitly declare postgres requirement for module and test

labkey-tchad:
Should clear up failures when running tests on SQL Server.

Verifying on TeamCity: https://teamcity.labkey.org/viewQueued.html?itemId=885437


labkey-jeckels (approved):

vagisha (approved):
Looks good to me.  Thanks!
I think MacCossLabModules/passport/module.properties should also have SupportedDatabases: pgsql since I don't see sqlserver scripts.  And PassportTest should also implement PostgresOnlyTest. Although I  don't think that gets run.

vagisha:
I committed changes to the passport module that I mentioned in the comment above.  I hope that is alright.  SchemaXMLTestCase failed in the last run on TeamCity.

vagisha:
Tests are passing on TeamCity: https://teamcity.labkey.org/viewLog.html?buildId=885540&buildTypeId=LabKey_193ReleaseSnapshot_Community_ExternalModulesTestSqlserver
There is one failure but that is not related to MacCossLabModules.

PR #24, opened 2020-01-20
LINCS: Remove old code related to generating processed (L4) GCTs

vagisha:
Removed old code related to generating processed (L4) GCTs on PanoramaWeb. Processed GCTs (L3, L4) are being generated on the Broad Institute's server and uploaded to PanoramaWeb.

Added a module property to set the assay type for a container.  The assay type is used to decide which R script will be run to generate the L2 GCT on PanoramaWeb, as well as sent as parameter to the Broad server.


labkey-jeckels (approved):
I'm not familiar with this code, but did a skim of the change and nothing popped up as being a problem. I'm all in favor of deleting unnecessary code!

PR #28, opened 2020-02-04
PanoramaPublic updates

vagisha:
1. Add an option to select the data license (CC BY 4.0 or CC0 for now) when submitting to Panorama Public.
2. Add a "Public" column to ExperimentAnnotationsTable so that it is easy to tell if an experiment is public or not.
3. Store the lab head details (name, email and affiliation) in JournalExperiment.  This will be used if the lab head does not have an account on the server.
4. Experiment title has to be >= 30 characters per ProteomeXchange.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/model/DataLicense.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/model/DataLicense.java
@@ -1,0 +1,65 @@
+package org.labkey.panoramapublic.model;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.Parameter;
+import org.labkey.api.util.Link;
+
+public enum DataLicense implements Parameter.JdbcParameterValue
+{
+    CC_BY_4_0 ("CC BY 4.0", "https://creativecommons.org/licenses/by/4.0/"),
+    CC0_1_0 ("CC0 1.0", "https://creativecommons.org/publicdomain/zero/1.0/");
+
+    private final String _displayName;
+    private final String _url;
+
+    DataLicense(String displayName, String url)
+    {
+        _displayName = displayName;
+        _url = url;
+    }
+
+    public String getDisplayName()
+    {
+        return _displayName;
+    }
+
+    public String getUrl()
+    {
+        return _url;
+    }
+
+    public Link getDisplayLink()
+    {
+        return new Link.LinkBuilder(getDisplayName()).href(getUrl()).target("_blank").clearClasses().build();
+    }
+
+    public String getDisplayLinkHtml()
+    {
+        return getDisplayLink().getHtmlString().toString();
+    }
+
+    public static DataLicense defaultLicense()
+    {
+        return DataLicense.CC_BY_4_0;
+    }
+
+    public static DataLicense resolveLicense(String license)
+    {
+        DataLicense resolved = DataLicense.valueOf(license);
+        return resolved == null ? defaultLicense() : resolved;
+    }
+
+    @Override
+    public @Nullable Object getJdbcParameterValue()
+    {
+        return name();
+    }
+
+    @Override
+    public @NotNull JdbcType getJdbcParameterType()
+    {
+        return JdbcType.VARCHAR;
+    }
+}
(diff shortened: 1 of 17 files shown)


panoramapublic/src/org/labkey/panoramapublic/proteomexchange/SubmissionDataValidator.java
```diff
@@ -113,6 +114,15 @@ public static SubmissionDataStatus validateExperiment(ExperimentAnnotations expA
     private static List<String> getMissingExperimentMetadataFields(ExperimentAnnotations expAnnot)
     {
         List<String> errors = new ArrayList<>();
+        if(StringUtils.isBlank(expAnnot.getTitle()))
+        {
+            errors.add("Title is required.");
+        }
+        else if(StringUtils.deleteWhitespace(expAnnot.getTitle()).length() < MIN_TITLE_LENGTH)
```
labkey-jeckels: This might be a little misleading in terms of an error message, since you're stripping whitespace from the middle, which I'd normally expect to count in the length. But maybe that's just because I'm a developer.

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
```diff
@@ -203,6 +204,67 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
         ExprColumn runCountColumn = new ExprColumn(this, "Runs", runCountSQL, JdbcType.INTEGER);
         addColumn(runCountColumn);
 
+        var isPublicCol = wrapColumn("Public", getRealTable().getColumn("Id"));
+        isPublicCol.setDisplayColumnFactory(colInfo -> new DataColumn(colInfo)
+        {
+            @Override
+            public Object getValue(RenderContext ctx)
+            {
+                Integer experimentAnnotationsId = ctx.get(FieldKey.fromParts(colInfo.getAlias()), Integer.class);
```
labkey-jeckels: Just use colInfo.getFieldKey(). The alias may or may not match its field key.

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
```diff
@@ -203,6 +204,67 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
         ExprColumn runCountColumn = new ExprColumn(this, "Runs", runCountSQL, JdbcType.INTEGER);
         addColumn(runCountColumn);
 
+        var isPublicCol = wrapColumn("Public", getRealTable().getColumn("Id"));
+        isPublicCol.setDisplayColumnFactory(colInfo -> new DataColumn(colInfo)
```
labkey-jeckels, line 208: Have you tested that these export the desired values to Excel and TSV, and that the client API returns the right values? The methods you override look good, but it's sometimes tricky to be sure you've covered all cases. You can also just decide that you don't care about those scenarios for these columns.

panoramapublic/src/org/labkey/panoramapublic/model/DataLicense.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public Link getDisplayLink()
+    {
+        return new Link.LinkBuilder(getDisplayName()).href(getUrl()).target("_blank").clearClasses().build();
+    }
+
+    public String getDisplayLinkHtml()
+    {
+        return getDisplayLink().getHtmlString().toString();
```
labkey-jeckels: Suggest making this return HtmlString to make it clearer who has been responsible for encoding.

labkey-jeckels (approved):
Some suggestions, but nothing mandatory.

panoramapublic/src/org/labkey/panoramapublic/proteomexchange/SubmissionDataValidator.java
vagisha: No, its not supposed to remove all white space! Thanks for catching!

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
vagisha: Done

panoramapublic/src/org/labkey/panoramapublic/model/DataLicense.java
vagisha: Done

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
vagisha: Done

PR #37, opened 2020-03-25
Adding PanoramaWeb Dashboard module

vagisha:
Contains queries and code to render PanoramaWeb usage charts.  
Creating PR against the right base branch!

### DIFF
--- a/dashboard/module.properties
+++ b/dashboard/module.properties
@@ -1,0 +1,8 @@
+Name: PanoramaWeb Dashboard
+ModuleClass: org.labkey.dashboard.DashboardModule
+ModuleDependencies: TargetedMS
+Label: PanoramaWeb Dashboard
+Description: Queries and code to render PanoramaWeb usage charts.
+URL: https://panoramaweb.org/wiki/home/page.view?name=Dashboard
+License: Apache 2.0
+LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
--- a/dashboard/src/org/labkey/dashboard/FolderDocsTable.java
+++ b/dashboard/src/org/labkey/dashboard/FolderDocsTable.java
@@ -1,0 +1,105 @@
+package org.labkey.dashboard;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.data.BaseColumnInfo;
+import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.ContainerForeignKey;
+import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.CoreSchema;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.WrappedColumn;
+import org.labkey.api.query.ExprColumn;
+import org.labkey.api.query.FieldKey;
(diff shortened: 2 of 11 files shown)


dashboard/src/org/labkey/dashboard/FolderDocsTable.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Created by vsharma on 12/8/2016.
+ */
+public class FolderDocsTable extends FilteredTable<UserSchema>
+{
+    public FolderDocsTable(@NotNull UserSchema userSchema, ContainerFilter cf)
+    {
+        super(CoreSchema.getInstance().getTableInfoContainers(), userSchema, cf);
+
+        setName("FolderDocs");
```
labkey-jeckels, line 30: Can you create a constant for this name and use it in the schema class too?

dashboard/module.properties
```diff
@@ -0,0 +1,8 @@
+Name: PanoramaWeb Dashboard
+ModuleClass: org.labkey.dashboard.DashboardModule
+ModuleDependencies: TargetedMS
+Label: PanoramaWeb Dashboard
```
labkey-jeckels, line 4: Please flag this module as Postgres only, since I don't see any need for SQLServer support and that will limit the need to test on both DBs

dashboard/src/org/labkey/dashboard/ProjectAdminsTable.java
```diff
@@ -0,0 +1,274 @@
...
+                    }
+                }
+            }
+            return allAdmins;
+        }
+
+        private String getUserDetails(Container container, UserPrincipal member)
+        {
+            ActionURL url = PageFlowUtil.urlProvider(UserUrls.class).getUserDetailsURL(container, member.getUserId(), null);
+            return new StringBuilder().append("<a href='").append(url.getLocalURIString()).append("'>").append(member.getName()).append("</a>").toString();
```
labkey-jeckels, line 259: Need to HTML-encode the name. Consider using our new-ish org.labkey.api.util.DOM builder utility

dashboard/module.properties
```diff
@@ -0,0 +1,8 @@
+Name: PanoramaWeb Dashboard
+ModuleClass: org.labkey.dashboard.DashboardModule
+ModuleDependencies: TargetedMS
```
labkey-jeckels, line 3: Dependencies should now be declared in the module's build.gradle. See the TargetedMS module's copy for an example

labkey-jeckels (changes requested):
I haven't tried running the code to test the behavior for myself, but what's the expectation in terms of what data should be shown to what users? I'm unsure what permission checks and/or filtering are required here, if any.

labkey-jeckels (changes requested):
@labkey-tchad made a good point that we should make the module name more specific here. Can you change the parent directory from "dashboard" to "panoramawebdashboard" or similar, since all module names share a global scope? Thanks!

labkey-jeckels:
Renaming the Java package and other similar scoping makes sense too to help avoid future conflicts

vagisha:
Created new PR with requested changes: https://github.com/LabKey/MacCossLabModules/pull/38

PR #38, opened 2020-03-27
Adding pwebdashboard module. 

vagisha:
Changes made after review:
- Renamed module (dashboard to pwebdashboard)
- Declared dependency on TargetedMS module in build.gradle
- Flagged module as Postgres only
- Other code changes

This module can only be added by site admins (PwebDashboardModule.getRequireSitePermission() returns true). It has been added in only one private folder on PanoramaWeb.  
There are no db tables.  The two tables added to the UserSchema both extend ContainerTable, and I confirmed that the container filter is applied correctly for the user's permissions. These tables are queried to render the usage charts in a Wiki page on PanoramaWeb.  The charts and a couple of other user-defined queries should eventually be added to this module.

### DIFF
--- a/pwebdashboard/build.gradle
+++ b/pwebdashboard/build.gradle
@@ -1,0 +1,6 @@
+import org.labkey.gradle.util.BuildUtils;
+
+dependencies {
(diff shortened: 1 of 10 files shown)


labkey-jeckels (approved):
Looks good, thank you!

PR #42, opened 2020-04-09
Check Bibliospec source files

kaipot:


vagisha:
If Josh approves merge into release20.3-SNAPSHOT for the targetedms changes then this should also be merged into the same branch.  I suspect your branch was created off release20.3-SNAPSHOT that is why you are seeing commits here (pwebdashboard module, and maccosslabmodules test suite) that are in release20.3-SNAPSHOT but not yet in develop.

panoramapublic/src/org/labkey/panoramapublic/proteomexchange/SubmissionDataStatus.java
```diff
@@ -109,6 +111,34 @@ public void addMissingRawPath(String filePath, String skylineDoc)
         skyDocs.add(skylineDoc);
     }
 
+    public Map<String, MissingLibrarySourceFiles> getMissingLibFiles()
+    {
+        return _missingLibFiles;
+    }
+
+    public void addMissingLibFile(String lib, String skylineDoc, String[] ssf, String[] idFiles)
```
labkey-jeckels: Similar question here - can these be lists instead of arrays? Looks like the calling code is converting a list to an array, which then gets converted back to a list.

labkey-jeckels (changes requested):

PR #48, opened 2020-05-28
Attempt to fix PassportTest

labkey-jeckels:
#### Changes
Fix login/logout workflow

Attempt to rearrange portal configuration to understand failure on TeamCity

Remove call to JS function that doesn't exist - caught in error check at end of test


labkey-jeckels:
@vagisha looks like this test hasn't yet passed on TeamCity, though the failure condition just changed.

https://teamcity.labkey.org/project.html?tab=testDetails&projectId=LabkeyTrunk&testNameId=6142058724692964943&page=2

labkey-jeckels:
Now passing on TeamCity with the portal config fixes in platform

labkey-ankurj (approved):

PR #55, opened 2020-06-18
20.3 fb panoramapublic submission and copy updates

vagisha:
#### Rationale
Panorama Public requests should be posted to a message board instead of being sent to an email address.  Creating and assigning a reviewer account and a ProteomeXchange ID should be part of the pipeline job that copies the user folder to Panorama Public.

#### Changes
* Added a site admin action to save the ProteomeXchange credentials.
* A SupportContainer column was added to the Journal table.  This is the container where Panorama Public requests are posted and can be set through the Admin console. 
* Requests to Panorama Public (create, edit, delete and re-submit) are posted to a message thread
    - AccouncementRowId column was added to the JournalExperiment table to keep track of the message thread.
* Changes to the copy pipeline job
    - Assign a PX ID to the data when requested
    - Create and assign a reviewer account to the copied data if a reviewer account was requested, otherwise make the folder public
    - Email the submitter with the PX ID, reviewer account etc. details after copying.  Post email contents to the message thread as well.


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicModule.java
```diff
@@ -65,7 +65,7 @@ public String getName()
     @Override
     public @Nullable Double getSchemaVersion()
     {
-        return 19.31;
+        return 20.32;
```
labkey-jeckels: It's totally fine to call this 20.32, but we've recently switched from versioning the schema to the LabKey release to the year that the script was added.

https://www.labkey.org/Documentation/wiki-page.view?name=sqlScripts#ver

There's absolutely no need to change this here, but wanted to make sure you saw the newer approach.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -538,6 +570,239 @@ public void setId(int id)
...
+            JspView view = new JspView<>("/org/labkey/panoramapublic/view/managePxCredentials.jsp", form, errors);
+            view.setFrame(WebPartView.FrameType.PORTAL);
+            view.setTitle("ProteomeXchange Credentials");
+            return view;
+        }
+
+        @Override
+        public NavTree appendNavTrail(NavTree root)
+        {
+            return null;
```
labkey-jeckels: Probably want to give this page a title

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -590,6 +860,16 @@ private void validateAction(CopyExperimentForm form)
...
             }
+            _journalExperiment = JournalManager.getJournalExperiment(_experiment.getId(), _journal.getId());
+            if(_journalExperiment == null)
+            {
+                throw new NotFoundException("Could not find an entry in JournalExperiment table for experimentId " + _experiment.getId()
+                + " and journalId " + _journal.getId());
+            }
+            if(_journalExperiment.getCopied() != null)
+            {
+                throw new UnsupportedOperationException(String.format("The experiment ID %d has already been copied.  It cannot be copied again.", _experiment.getId()));
```
labkey-jeckels: This will get reported to the mothership exception reporting. Not a big deal, but this looks like a data problem, not a code bug. Consider using BadRequestException instead, or something else that implements SkipMothershipLogging

labkey-jeckels (approved):
A few suggestions, but nothing mandatory

PR #96, opened 2020-12-03
Issue 1713: Suppress trailing ? in URL strings if no parameters

labkey-kevink:
#### Rationale
With the experimental no-question-mark-url flag enabled, it isn't safe to concatenate parameters to the end of a URL.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/1603
- https://github.com/LabKey/platform/pull/1997

#### Changes
- don't use string concatenation to add parameters to URLs


testresults/src/org/labkey/testresults/view/longTerm.jsp
```diff
@@ -109,9 +109,11 @@
                 });
                 $('#failureTable').empty();
                 var failureTableHTML = "<tr><td>Test</td><td>Failures per Run(%)</td></tr>";
+                let url = new URL(<%=q(new ActionURL(TestResultsController.ShowFailures.class, c).addParameter("viewType", "yr").getURIString())%>);
                 for(var i = 0; i < failuresByFailToRunRatio.length; i++) {
                     var failure = failuresByFailToRunRatio[i];
-                    failureTableHTML += '<tr><td><a href="<%=h(new ActionURL(TestResultsController.ShowFailures.class, c))%>viewType=yr&failedTest='+failure[0]+'">'+failure[0]+'</a></td><td>'+failure[1]+'</td></tr>';
+                    url.searchParams.set('failedTest', failure[0]);
+                    failureTableHTML += '<tr><td><a href="' + url.toString() + '">'+failure[0]+'</a></td><td>'+failure[1]+'</td></tr>';
```
labkey-adam: HTML encode url and failures?

labkey-adam:
Should we introduce a new `JspBase.jsUrl()` method to simplify this pattern? I'm thinking:

```<%=jsUrl(new ActionURL(ShowRunAction.class, c))%>```

instead of

```new URL(<%=q(new ActionURL(ShowRunAction.class, c).getURIString())%>)```

This continues the theme of using ActionURL as much as possible, reducing the number of `getURIString()` references to check/annotate.

**Review by labkey-adam — CHANGES_REQUESTED** [review]
Couple things to consider

File: testresults/src/org/labkey/testresults/view/longTerm.jsp (hunk 1 above)
Comment (labkey-kevink) [reply testresults/src/org/labkey/testresults/view/longTerm.jsp]: Fixed

**Comment by labkey-kevink** [comment]
I added the `jsURL()` method and it works, but IntelliJ doesn't give good code completion for the local url variable when we do this.  We can add a local type annotation to the block, but it gets a little ugly:

```
/** @type URL */
var url = <%=jsURL(new ActionURL(...))%>;
// IntelliJ can code complete the properties on the url variable now
```

labkey-adam:
Looks okay to me. Most (all?) of these could be simplified by using `urlFor(actionClass)`. Even better... another variant of your new method: `jsUrl(actionClass)`. Optional... but super cool :)

labkey-adam (approved):

PR #110, opened 2021-03-06
Wait for page loads

labkey-tchad:
#### Rationale
When a test does something that triggers a page load, it needs to wait for the navigation. Otherwise, it's possible for subsequent actions to fail because the test isn't on the page they expect. I noticed some intermittent failures on TeamCity in the `PanoramaPublicTest` due to this.

#### Changes
* Use `clickAndWait` to handle navigation properly

### DIFF
--- a/panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaPublicTest.java
+++ b/panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaPublicTest.java
@@ -115,7 +115,7 @@ public class PanoramaPublicTest extends TargetedMSTest implements PostgresOnlyTe
         testSubmitWithMissingRawFiles(portalHelper, expWebPart);
 
         // Submit the experiment by clicking the "Continue without a ProteomeXchange ID" link
-        portalHelper.click(Locator.folderTab("Panorama Dashboard"));
+        clickAndWait(Locator.folderTab("Panorama Dashboard"));
         expWebPart.submitWithoutPXId();
         assertTextPresent("Copy Pending!");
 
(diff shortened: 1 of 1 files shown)


vagisha (approved):
Thanks!

labkey-sweta (approved):

panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaPublicTest.java
```diff
@@ -291,12 +291,12 @@ public DataRegionTable getDataRegion()
 
         public void submitWithoutPXId()
         {
-            findElement(Locator.linkContainingText("Submit")).click();
-            waitAndClick(Locator.linkContainingText("Continue without a ProteomeXchange ID"));
+            clickAndWait(Locator.linkContainingText("Submit"));
```
labkey-danield, line 294: Does this need to be a `waitAndClickAndWait`?

panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaPublicTest.java
```diff
@@ -291,12 +291,12 @@ public DataRegionTable getDataRegion()
...
+            clickAndWait(Locator.linkContainingText("Continue without a ProteomeXchange ID"));
             getWrapper()._ext4Helper.selectComboBoxItem(Ext4Helper.Locators.formItemWithInputNamed("journalId"), PANORAMA_PUBLIC);
-            waitAndClick(Ext4Helper.Locators.ext4Button("Submit"));
-            waitAndClick(Locator.lkButton("OK")); // Confirm to proceed with the submission.
-            waitAndClick(Locator.linkWithText("Back to Experiment Details")); // Navigate to the experiment details page.
+            clickAndWait(Ext4Helper.Locators.ext4Button("Submit"));
```
labkey-danield, line 297: I'm guessing these don't need to be a `waitAndClickAndWait` because the action before this was a `clickAndWait`. Correct?

PR #117, opened 2021-04-04
Issue 42801: Ask user if they want to include subfolders in data submitted to Panorama Public

vagisha:
#### Rationale
It is possible for a user to configure an experiment submitted to Panorama Public to include subfolders. However, the option is hard to find, and it is not obvious to the user if subfolders will be included in the copy made to Panorama Public. If there are subfolders, we want the user to confirm that they want to include (or exclude) them when they submit an experiment copy request to Panorama Public

#### Changes
- When a user submits a request to Panorama Public, and the experiment is not configured to include subfolders, then the user will be presented options to either include subfolders or submit just the folder without subfolders
- Submitting user is required to have read permissions in all subfolders if they want to include them in the submission request.
- look for ReadPermission instead of InsertPermission when including subfolders in the experiment.
- Update the SecurityPolicy on the ShortURL if the user resubmitting an experiment is different from the one who originally submitted it


labkey-jeckels (approved):
Looks good to me.

PR #126, opened 2021-07-28
Issue 42942: Support chromatogram libraries in Panorama Public

vagisha:
#### Rationale
Panorama Public currently only supports experimental data folders. Add support for submitting and copying chromatogram library folders.

#### Changes
- Added UpdateFolderTypeTask to the panoramapublic pipeline that will copy the library state of runs, peptide groups, and precursors from the source library folder.
- Library state of entities in the source folder is exported as a tab-separated file to '@files/targetedMSLib' in the target folder.  Library state of entities in the target folder are updated using this file
- Add a folder property in the target folder with the chromatogram library revision
- Rename the copied .clib files with the target container rowId so that they are recognized in the target container
- Added tests

PR #130, opened 2021-09-06
Issue 43864: Add support for dataset versioning in Panorama Public

vagisha:
#### Rationale
We’ve recently added support for publishing chromatogram libraries to Panorama Public.  Since we expect chromatogram libraries to evolve over time, we need a mechanism to have more than one published version of a library on Panorama Public. 
The data versioning implemented in this PR will work for both library and experimental data folders.  It is described in the doc linked here: https://www.labkey.org/MacCoss/Issue%20Tracker/issues-details.view?issueId=43864

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/126

#### Changes
- Schema changes: 
  -  added a 'Submission' table populated with rows from the JournalExperiment table
  - added a 'DataVersion' column to the 'ExperimentAnnotations' table
- Show version in the TargetedMSExperiment webpart of the Panorama Public copy if there are more than one copied versions of an experiment
- Added a 'Published Versions' table on the experiment details page in the user's folders; 'Submission' table shows a record of all the submission requests
- Added 'Version' and 'Version Count' columns in the ExperimentAnnotationsTableInfo (displayed on the home page of Panorama Public). They are not included the list of default columns
- Added action class to view published versions of an experiment
- Always allow 'resubmit' of 'Library' folders. This is different from 'Experimental Data' folders which cannot be resubmitted after a paper is published
- JournalSubmission class that encapsulates a JournalExperiment and associated Submission rows
- SubmissionManager class added for querying and updating submissions
- Clean up of CopyExperimentFinalTask and action classes in PanoramaPublicController impacted by the change.
- Updated PanoramaPublicTest

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/model/JournalSubmission.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/model/JournalSubmission.java
@@ -1,0 +1,161 @@
+package org.labkey.panoramapublic.model;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.view.ShortURLRecord;
+import org.labkey.panoramapublic.query.SubmissionManager;
+
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class JournalSubmission
+{
+    private final JournalExperiment _journalExperiment;
+    private List<Submission> _submissions;
+
+    public JournalSubmission(@NotNull JournalExperiment journalExperiment)
+    {
+        _journalExperiment = journalExperiment;
+    }
+
+    public JournalExperiment getJournalExperiment()
+    {
+        return _journalExperiment;
+    }
+
+    public int getJournalExperimentId()
+    {
+        return _journalExperiment.getId();
+    }
+
+    public int getJournalId()
+    {
+        return _journalExperiment.getJournalId();
+    }
+
+    public int getExperimentAnnotationsId()
+    {
+        return _journalExperiment.getExperimentAnnotationsId();
+    }
+
+    public ShortURLRecord getShortAccessUrl()
+    {
+        return _journalExperiment.getShortAccessUrl();
+    }
+
+    public ShortURLRecord getShortCopyUrl()
+    {
+        return _journalExperiment.getShortCopyUrl();
+    }
+
+    public Date getCreated()
+    {
+        return _journalExperiment.getCreated();
+    }
+
+    public int getCreatedBy()
+    {
+        return _journalExperiment.getCreatedBy();
+    }
+
+    public Date getModified()
+    {
+        return _journalExperiment.getModified();
+    }
+
+    public int getModifiedBy()
+    {
+        return +_journalExperiment.getModifiedBy();
+    }
+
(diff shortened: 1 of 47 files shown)


panoramapublic/src/org/labkey/panoramapublic/model/JournalSubmission.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public Date getModified()
+    {
+        return _journalExperiment.getModified();
+    }
+
+    public int getModifiedBy()
+    {
+        return +_journalExperiment.getModifiedBy();
```
labkey-jeckels: What's this extra plus doing here?

panoramapublic/src/org/labkey/panoramapublic/query/SubmissionManager.java
```diff
@@ -0,0 +1,413 @@
...
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class SubmissionManager
+{
+    private static final Logger LOG = LogManager.getLogger(SubmissionManager.class);
+
+    public static Submission getSubmission(int id)
```
labkey-jeckels: I worry that many of these methods that take an ID are subject to direct object reference. I spot checked a few of them and it looks like the callers are not validating the owning container. So if you can guess the ID for a row you don't have access to (not that hard since they're auto-incrementing) you may be able to view or manipulate it.

To fix it, you'd pass in a Container (if you know the container it'll be in), or a User and make sure that the user has at least read permission in the object's container.

https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html

labkey-jeckels (changes requested):
Overall looks good but see my comment about direct object reference security concerns.

panoramapublic/src/org/labkey/panoramapublic/model/JournalSubmission.java
vagisha: No idea!  Thanks for catching!

panoramapublic/src/org/labkey/panoramapublic/query/SubmissionManager.java
vagisha: Actions that operate on a submission request need a valid ExperimentAnnotations object.  And all of the relevant action classes in PanoramaPublicController call the ensureCorrectContainer(Container requestContainer, Container expAnnotContainer, ViewContext viewContext) method that throws a RedirectException if the container in the request is not the same as the owning container of the ExperimentAnnotations object. 
So it would not be possible for a user to edit/update/delete a submission request by using the row Id for a submission that belonged to an experiment in a container where they do not have the required permission. 

But I've also added a Container parameter to SubmissionManager methods that take a database row ID, as you suggested.

panoramapublic/src/org/labkey/panoramapublic/model/JournalSubmission.java
labkey-jeckels: I was surprised that was valid syntax!

labkey-jeckels (approved):

PR #135, opened 2021-10-30
Allow users to make their data public

vagisha:
- Add a "Make Public" button to the TargetedMSExperiment webpart that will let users make their data on Panorama Public public.  This button will be displayed in the user's copy of the data, in their project. This will:
  - Display a form where the user can fill the publication details (PubMed ID or link and citation)
  - Assign Reader role to Site:Guests and the user account that can be used to download files in a WebDAV client
  - Add a "Download Data" webpart to the "Raw Data" tab.  This webpart has the WebDAV URL for the "RawFiles" folder in the file root along with the user email and password that can be used to connect to the folder in a WebDAV client
  - If a PubMed ID was entered, we will get the citation in NLM format through PubMed's Literature Citation Exporter
  - Set the publication details on the ExperimentAnnotations in the Panorama Public folder for the data
  - Make the DOI "findable" if one was assigned
- AddPublicDataUserAction class was added to let site admins set the user account that should be used for WebDAV downloads:  (Admin Console > Panorama Public > Add Public Data User)

 Changes are described here: https://docs.google.com/document/d/1cjLH-OBc77jHBn_DN4u9munUl7EA_e2RwYfIqWjPcFM/edit

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
@@ -19,7 +19,6 @@ import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.validator.routines.EmailValidator;
 import org.apache.commons.validator.routines.UrlValidator;
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -101,6 +100,7 @@ import org.labkey.api.security.permissions.InsertPermission;
 import org.labkey.api.security.permissions.ReadPermission;
 import org.labkey.api.security.permissions.UpdatePermission;
 import org.labkey.api.security.roles.ProjectAdminRole;
+import org.labkey.api.security.roles.ReaderRole;
 import org.labkey.api.targetedms.TargetedMSService;
 import org.labkey.api.util.Button;
 import org.labkey.api.util.DOM;
@@ -109,8 +109,10 @@ import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.HtmlString;
 import org.labkey.api.util.Link;
 import org.labkey.api.util.PageFlowUtil;
+import org.labkey.api.util.Pair;
 import org.labkey.api.util.TestContext;
 import org.labkey.api.util.URLHelper;
+import org.labkey.api.util.logging.LogHelper;
 import org.labkey.api.view.*;
 import org.labkey.api.view.template.ClientDependency;
 import org.labkey.panoramapublic.chromlib.ChromLibStateManager;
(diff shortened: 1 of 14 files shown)


labkey-adam:
Sounds cool, @vagisha !

panoramapublic/src/org/labkey/panoramapublic/view/publish/dataDownloadInfo.jsp
```diff
@@ -0,0 +1,32 @@
...
+<%@ page import="org.labkey.api.util.PageFlowUtil" %>
+<%@ page import="org.labkey.api.settings.AppProps" %>
+<%@ page extends="org.labkey.api.jsp.JspBase" %>
+<%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
+<labkey:errors/>
+<%
+    JspView<JournalManager.PublicDataUser> me = (JspView<JournalManager.PublicDataUser>) HttpView.currentView();
+    var publicDataUser = me.getModelBean();
+    // NOTE: This is a link to the data download documentation page on PanoramaWeb.  It will not work on any other server.
+    var downloadDataDocHref = "/wiki/home/page.view?name=download_public_data";
```
labkey-jeckels: Suggest using the newer URL format:
```suggestion
    var downloadDataDocHref = "/home/wiki-page.view?name=download_public_data";
```

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -725,6 +738,138 @@ public Container getSupportContainer()
...
+                {
+                    form.setUserEmail(publicDataUser.getEmail());
+                }
+            }
+
+            HtmlView view = new HtmlView(
+                    DIV(
+                            ERRORS(errors),
+                            "Enter the email address for the user account that can be used to download public datasets in a WebDAV client",
+                            FORM(at(method, "POST", action, new ActionURL(AddPublicDataUserAction.class, getContainer()).addParameter("id", _journal.getId())),
```
labkey-jeckels, line 789: Why is it important to let the admin choose their own username/password here? Couldn't everyone use the public@proteinms.net account?

labkey-jeckels (approved):
A couple of comments but nothing that needs to hold a merge.

This PR may just miss the cutoff for the 21.11 branch, so you way want to adjust the target branch once it's been created. Since develop and the release branch haven't diverged, we shouldn't need to separately run tests and such on the new branch.

panoramapublic/src/org/labkey/panoramapublic/view/publish/dataDownloadInfo.jsp
vagisha: Fixed. Thanks for catching.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: This is a one-time setup for a site admin to choose the user account that can be used for downloading files in a WebDAV client.  The user Id and password are saved in an encrypted store, and are used when rendering dataDownloadInfo.jsp.

labkey-jeckels: Got it. Makes sense. I was thinking this was being configured at the per-folder level.

PR #136, opened 2021-11-01
Auto-populate "Instrument" field; include first and last name in "Authors" column

vagisha:
- Auto populate the "Instrument" field in the new ExperimentAnnotations form.
  - Get the instrument names from the Skyline document metadata (targetedms.instrument table) in the container.
  - Instrument model names must match instruments in the PSI-MS CV. This should work for Thermo and SCIEX instrument but usually not for other instrument vendors.
- Look up members of groups with folder admin role in the container to populate the drop down menu for the "Submitter" and "Lab Head" fields in the form.
  This is in addition to users with folder admin roles.
- Include the first and last names of submitter and lab head instead of the display name in the "Authors" column (related to PR 131).

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
@@ -4008,6 +4008,11 @@ public class PanoramaPublicController extends SpringActionController
             addExperimentViewDependencies(view);
             view.setTitle(TargetedMSExperimentWebPart.WEB_PART_NAME);
             view.setInitialValue(SUBMITTER, getUser().getUserId());
+            List<PsiInstrumentParser.PsiInstrument> instruments = ExperimentAnnotationsManager.getContainerInstruments(getContainer(), getUser());
+            if (instruments.size() > 0)
+            {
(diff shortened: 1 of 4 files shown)


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -4008,6 +4008,11 @@ public ModelAndView getView(NewExperimentAnnotationsForm form, BindException err
             addExperimentViewDependencies(view);
             view.setTitle(TargetedMSExperimentWebPart.WEB_PART_NAME);
             view.setInitialValue(SUBMITTER, getUser().getUserId());
+            List<PsiInstrumentParser.PsiInstrument> instruments = ExperimentAnnotationsManager.getContainerInstruments(getContainer(), getUser());
+            if (instruments.size() > 0)
+            {
+                view.setInitialValue("instrument", StringUtils.join(instruments.stream().map(i -> i.getName()).collect(Collectors.toList()), ","));
```
labkey-jeckels, line 4014: I assume the name is never null?

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: Yes, that is correct.

PR #138, opened 2021-11-04
Copy ExpData comments (descriptions) to the Panorama Public copy 

vagisha:
#### Rationale
FolderWriter / FolderImporter do not copy the comments (displayed in the Description column of the Files browser) entered by a user for files / folders.  

#### Changes
Added FilesMetadataWriter that gets called in ExperimentExportTask after the folder export is complete.  FilesMetadataWriter writes one files.xml for every exported folder (the root folder and any children if they are included in the export). This file contains, for each ExpData in the folder that has a comment, the path relative to the container file root and the comment.
FilesMetadataImporter is called at the end of ExperimentImportTask after FolderImporter is done.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/pipeline/ExperimentExportTask.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/pipeline/ExperimentExportTask.java
@@ -15,7 +15,6 @@
  */
 package org.labkey.panoramapublic.pipeline;
 
-import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.admin.FolderExportContext;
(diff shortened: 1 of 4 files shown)


labkey-jeckels (commented):
Overall, this looks good to me. It doesn't yet handle customized file properties, but it's great to capture the comments. I think we should move this into FileWriter and FileImporter though. It looks like it should be fairly straightforward, and would simplify the new code because it would no longer need to handle the recursion into subfolders on its own.

My only question is whether we'd want to use a different XML schema or other file format to make it easier to support the full set of properties in the future. They could certainly be added to this XML format, but it would be nice to avoid inventing new variants of metadata definitions and property values when possible. @labkey-adam any recommendations?

labkey-jeckels (approved):
Also, this PR currently targeting develop. Vagisha, did you want this for 21.11? I'll go ahead and approve this PR as-is, because we can always remove this copy of the code whenever we add this functionality to FileWriter and FileImporter, in 21.11 or later.

labkey-adam:
Yes, ideally we would export/import using `TableType`, perhaps exporting the schema using `TableInfoWriter` (there's a variant `PropertyTableWriter` that might be worth glancing at) and importing via `ImportTypesHelper`. And then handling the data using `TSVGridWriter` and `TabLoader`. Or something like that...

vagisha:
It would be great if I can merge this with the 21.11 branch.  I can change the target if it is still OK to merge as is. This functionality can be added later to FileWriter and FileImporter using `TableType` as Adam suggests.

labkey-jeckels:
Yes, that should be fine. The tests are happy here, and develop and 21.11 are very similar at the moment. You should be able to edit this PR to change the target branch and then merge.

labkey-adam:
Of course, if the serialization approach changes then archives exported from 21.11 won't import with descriptions in a future version. Unless we keep the old import code in place for backward compatibility.

labkey-jeckels:
Yes, but for this use case the export is immediately imported, so it shouldn't be a big deal to change it.

labkey-adam:
Cool, didn't realize that

PR #155, opened 2022-01-04
Fix bug where activating/deactivating user did not work

kaipot:
#### Rationale
Activating/deactivating user did not work.


labkey-jeckels (approved):
I don't know this code or how to test it, but the change seems perfectly reasonable

PR #163, opened 2022-01-15
Issue 44679: Handle HTML tags in citations returned by Literature Citation Exporter

vagisha:
#### Rationale
Citations returned by the NCBI Literature Citation Exporter can contain HTML tags (e.g. `<sup>`, `<sub>`, `<i>`) for formatting. Pass the citation through `PageFlowUtil.validateHtml()` and `PageFlowUtil.sanitizeHtml()` and so that we can safely display the formatted citation.


labkey-jeckels (approved):
It looks like you were able to get it to catch <script> tags?

vagisha:
> It looks like you were able to get it to catch <script> tags?

Yes.  Calling `validateHtml()` with false as the `scriptAsErrors` argument catches `<script>` tags.

PR #176, opened 2022-03-03
Fix user schema for testresults

kaipot:
#### Rationale
Update to fix testresults schema, change DeleteRunAction and FlagRunAction to inherit from MutatingApiAction

### DIFF
--- a/testresults/src/org/labkey/testresults/TestResultsController.java
+++ b/testresults/src/org/labkey/testresults/TestResultsController.java
@@ -672,11 +672,12 @@ public class TestResultsController extends SpringActionController
      * action for deleting a run ex:'deleteRun.view?runId=x'
      */
     @RequiresPermission(AdminPermission.class)
-    public class DeleteRunAction extends BeginAction {
-
+    public static class DeleteRunAction extends MutatingApiAction {
         @Override
-        public ModelAndView getView(Object o, BindException errors) throws Exception
+        public Object execute(Object o, BindException errors)
         {
+            ApiSimpleResponse response = new ApiSimpleResponse();
+
             int rowId = Integer.parseInt(getViewContext().getRequest().getParameter("runId"));
             SimpleFilter filter = new SimpleFilter();
             filter.addCondition(FieldKey.fromParts("testrunid"), rowId);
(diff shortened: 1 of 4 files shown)


testresults/src/org/labkey/testresults/TestResultsController.java
```diff
@@ -687,22 +688,23 @@ public ModelAndView getView(Object o, BindException errors) throws Exception
                 Table.delete(TestResultsSchema.getTableInfoHangs(), filter);
                 Table.delete(TestResultsSchema.getTableInfoTestRuns(), rowId); // delete run last because of foreign key
                 transaction.commit();
+            } catch (Exception x) {
+                response.put("success", false);
```
labkey-jeckels, line 692: Logging this on the server side can sometimes be helpful to capture the full error info, including its stack trace. But this really depends on what types of exceptions you might expect

labkey-jeckels (approved):
A small, optional suggestion

testresults/src/org/labkey/testresults/view/runDetail.jsp
```diff
@@ -68,8 +68,7 @@
     <!--Header content, username run Id, and other information about the run-->
     <h2>
         <%=h(run.getUserName() + " : " + df.format(run.getPostTime()))%> &nbsp;
-        <img id="flagged" flagged="false"
-            src="<%=h(contextPath)%>/TestResults/img/<% if (run.isFlagged()) { %>flagon<% } else { %>flagoff<% } %>.png"
+        <img id="flagged" src="<%=h(contextPath)%>/TestResults/img/<% if (run.isFlagged()) { %>flagon<% } else { %>flagoff<% } %>.png"
```
vagisha: Maybe replace this with
src="<%=getWebappURL("/TestResults/img/" + (run.isFlagged() ? "flagon" : "flagoff" + ".png"))%>"

vagisha (approved):

labkey-tchad:
@labkey-jeckels @vagisha @kaipot 
I think the `testresults.xml` was fine the way it was. It is failing the [SchemaXMLTest now](https://teamcity.labkey.org/buildConfiguration/LabKey_217Release_Community_DrtPostgres/1732232?buildTab=tests&expandedTest=build%3A%28id%3A1732232%29%2Cid%3A4516).
We noticed the error when merging 21.7 to 21.11 today.

The [upgrade script](https://github.com/LabKey/MacCossLabModules/blob/develop/testresults/resources/schemas/dbscripts/postgresql/testresults-13.29-13.30.sql) that moved the `active` column from `user` to `userdata` is commented out.

labkey-jeckels:
@labkey-tchad @kaipot @labkey-sweta as far as TeamCity goes, that's correct. However, looking at skyline.ms (the server where this module is used), I believe that the script was probably run manually against the DB. It doesn't have a testresults.user.active column and does have a testresults.userdata.active column.

Thus, I suggest that we fix the tests by uncommenting the script. This won't affect skyline.ms as its schema version is already at 13.38.

vagisha:
Thanks, Josh.  Yes, that script must have been run manually on skyline.ms. Going forward, we will always use the installer builds from TeamCity for upgrading skyline.ms so that we don't see such discrepancies.

labkey-jeckels:
https://github.com/LabKey/MacCossLabModules/pull/178

PR #181, opened 2022-03-22
Only link tests in rundown table if there is a failure

kaipot:
#### Rationale
Tests in the rundown table should only be linked to the test failure page if they had a failure because if there is only a leak or hang, there are no failures to show.

### DIFF
--- a/testresults/src/org/labkey/testresults/view/rundown.jsp
+++ b/testresults/src/org/labkey/testresults/view/rundown.jsp
@@ -257,9 +257,9 @@
                         <tr>
                             <td style="width: 200px; max-width: 200px; overflow: hidden; text-overflow: ellipsis; padding: 0;">
                                 <% if (Arrays.stream(problemRuns).anyMatch(run -> problems.isFail(run, test))) { %>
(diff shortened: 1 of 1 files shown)


vagisha (approved):

testresults/src/org/labkey/testresults/view/rundown.jsp
```diff
@@ -255,7 +256,11 @@
                         <% for (String test : problemTests) { %>
                         <tr>
                             <td style="width: 200px; max-width: 200px; overflow: hidden; text-overflow: ellipsis; padding: 0;">
+                                <% if (Arrays.stream(problemRuns).anyMatch(run -> problems.isFail(run, test))) { %>
                                 <a href="<%=h(new ActionURL(TestResultsController.ShowFailures.class, c).addParameter("end", df.format(selectedDate)).addParameter("failedTest", test))%>" target="_blank"><%=h(test)%></a>
```
vagisha: You could replace this with
`<%=link(test).href(new ActionURL(TestResultsController.ShowFailures.class, c).addParameter("end", df.format(selectedDate)).addParameter("failedTest", test)).target("_blank")%>`
Add `.clearClasses()` if you don't like the default link style.

kaipot: Thanks, I changed it.

PR #175, opened 2022-02-28
Issue 44682: Run PX data validation as a pipeline job

vagisha:
#### Rationale
We would like to run the data validation step of submitting a dataset to Panorama Public as a pipeline job.  The details are described in this doc: https://docs.google.com/document/d/1zDJY5f8rbb1Nq95__JxFStVmPqtcufFkVV3FY_wNNw0/edit

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/497

#### Changes
- Run data validation for ProteomeXchange as a pipeline job
- Results from running the job are stored in new tables added to the panoramapublic schema
- PxValidationStatusApiAction returns the validation results as JSON which is displayed in pxValidationStatus.jsp
- CopyExperimentAction, PX actions (validate / submit / update PX XML), and PX XML and HTML writers have been updated to use the new data validation classes
- A spectral library is marked as incomplete if:
  - the SpectrumSourceFiles table of the .blib does not have the idFileName column, or there are now rows in the SpectrumSourceFiles table
  - the .blib library was built with an assay library (.csv files as source)
  - library is unsupported (only .blib and .elib are supported)
  - the library is referenced in the Skyline XML but not included in the .sky.zip
- For .blib built with MaxQuant search results, look for mqpar.xml etc. in addition to what is listed in the SpectrumSourceFiles.idFileName column
- If submitted data is in sub-folders we look for spectral library source files in the parent experiment folder as well as the sub-folders that have a Skyline document with the library. The files have to be uploaded to only one folder
- Sample files in a Skyline document that have the same file name but were imported from different paths will be marked as ambiguous, and the dataset will not be considered a valid PX dataset.  We may support this later by requiring that such files be uploaded into subfolders named for the replicate so that we can can link to the correct sample file.
- Added some fixes to matching modifications to Unimod in the case that a modification in the Skyline document does not have a Unimod ID. Updated test in ExperimentModificationGetter with modifications found on PanoramaWeb without a Unimod Id.
-  ExperimentUserForeignKey extends UserIdForeignKey so that the Submitter and LabHead columns can include all users when the experiment list is viewed in the Panorama Public project.
- Updated existing tests and added a PanoramaPublicValidationTest
- Moved test sample data to TargetedMS/panoramapublic subfolder

PR #184, opened 2022-03-24
Issue 44682: Run PX data validation as a pipeline job

vagisha:
New PR after renaming branch to target release22.3-SNAPSHOT.  Original PR: https://github.com/LabKey/MacCossLabModules/pull/175

#### Rationale
We would like to run the data validation step of submitting a dataset to Panorama Public as a pipeline job.  The details are described in this doc: https://docs.google.com/document/d/1zDJY5f8rbb1Nq95__JxFStVmPqtcufFkVV3FY_wNNw0/edit

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/497

#### Changes
- Run data validation for ProteomeXchange as a pipeline job
- Results from running the job are stored in new tables added to the panoramapublic schema
- PxValidationStatusApiAction returns the validation results as JSON which is displayed in pxValidationStatus.jsp
- CopyExperimentAction, PX actions (validate / submit / update PX XML), and PX XML and HTML writers have been updated to use the new data validation classes
- A spectral library is marked as incomplete if:
  - the SpectrumSourceFiles table of the .blib does not have the idFileName column, or there are now rows in the SpectrumSourceFiles table
  - the .blib library was built with an assay library (.csv files as source)
  - library is unsupported (only .blib and .elib are supported)
  - the library is referenced in the Skyline XML but not included in the .sky.zip
- For .blib built with MaxQuant search results, look for mqpar.xml etc. in addition to what is listed in the SpectrumSourceFiles.idFileName column
- If submitted data is in sub-folders we look for spectral library source files in the parent experiment folder as well as the sub-folders that have a Skyline document with the library. The files have to be uploaded to only one folder
- Sample files in a Skyline document that have the same file name but were imported from different paths will be marked as ambiguous, and the dataset will not be considered a valid PX dataset.  We may support this later by requiring that such files be uploaded into subfolders named for the replicate so that we can can link to the correct sample file.
- Added some fixes to matching modifications to Unimod in the case that a modification in the Skyline document does not have a Unimod ID. Updated test in ExperimentModificationGetter with modifications found on PanoramaWeb without a Unimod Id.
-  ExperimentUserForeignKey extends UserIdForeignKey so that the Submitter and LabHead columns can include all users when the experiment list is viewed in the Panorama Public project.
- Updated existing tests and added a PanoramaPublicValidationTest
- Moved test sample data to TargetedMS/panoramapublic subfolder


panoramapublic/resources/schemas/dbscripts/postgresql/panoramapublic-22.000-22.001.sql
```diff
@@ -0,0 +1,141 @@
...
+CREATE TABLE panoramapublic.SkylineDocValidation
+(
+    Id                         SERIAL NOT NULL,
+    ValidationId               INT NOT NULL,
+    RunId                      BIGINT NOT NULL, -- targetedms.runs.Id
+    Container                  ENTITYID NOT NULL, -- Container where the run lives. This could be a subfolder of the main experiment folder
+    Name                       VARCHAR(300) NOT NULL, -- Name of the Skyline file
+
+    CONSTRAINT PK_SkylineDocValidation PRIMARY KEY (Id),
+    CONSTRAINT FK_SkylineDocValidation_DataValidation FOREIGN KEY (ValidationId) REFERENCES panoramapublic.DataValidation(Id)
```
labkey-jeckels, line 44: Should probably have a FK on container too

labkey-jeckels (approved):

panoramapublic/resources/schemas/dbscripts/postgresql/panoramapublic-22.000-22.001.sql
vagisha: I removed the container column.  Getting the container from the targetedms run instead.

panoramapublic/src/org/labkey/panoramapublic/view/publish/pxValidationStatus.jsp
```diff
@@ -638,7 +636,7 @@
             });
             if (hasInferred) {
                 var noteHtml = "Modification names starting with <strong>**</strong> in the Name column did not have a Unimod Id in the Skyline document."
-                        + " A Unimod match was was inferred based on the formula, modification site(s) and terminus in the modification definition"
+                        + " A Unimod match was was assigned based on the formula, modification site(s) and terminus in the modification definition"
```
labkey-jeckels: Typo - "was was"

panoramapublic/src/org/labkey/panoramapublic/query/DataValidationManager.java
```diff
@@ -396,6 +394,182 @@ public static DataValidation updateDataValidation(DataValidation validation, Use
...
+    private static List<SpecLib> getLibrariesForSpecLibInfo(@NotNull SpecLibInfo specLibInfo, DataValidation validation)
+    {
+        List<SpecLib> specLibList = new ArrayList<>();
+
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("specLibInfoId"), specLibInfo.getId());
+        filter.addCondition(FieldKey.fromParts("validationId"), validation.getId());
+        List<SpecLib> specLibs = new TableSelector(PanoramaPublicManager.getTableInfoSpecLibValidation(), filter, null).getArrayList(SpecLib.class);
+        if (specLibs.size() > 0)
+        {
+            specLibList.addAll(specLibs);
```
labkey-jeckels: Is this needed? Could you just return specLibs directly and avoid the copy?

PR #188, opened 2022-03-28
Modification information interface

vagisha:
#### Rationale
Modifications in a Skyline document don't always have a Unimod Id. This can be due to modification definitions that combine two modifications, or because of reuse of older template documents or saved custom modifications.
For a "complete" ProteomeXchange submission all modifications in the submitted Skyline documents must have a Unimod Id. In the previous work (PR#184) an attempt was made to assign Unimod Ids during the data validation process. However, this can sometimes result in an incorrect assignment, especially for combination modifications.  We want to let the submitter assign Unimod Ids or define combination modifications.
The changes are described in this Google doc:
https://docs.google.com/document/d/1DcwjDDkQdV9ZspbRjFxLuxG0V1JREz4R26cOg2XJDII/edit

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/184

#### Changes
- Added an interface for users to:
  - Find a Unimod match for modifications that did not have a Unimod Id in the Skyline document .
  - Define a combination modification for a structural modification that is a combination of two modifications.
- Removed automatic Unimod matching from the ProteomeXchange DataValidator.
- Updated PX validation details UI to show user-assigned Unimod Ids and combination modifications.
- PX validation status is updated when a user-assigned modification information is added or deleted
- Modification information saved in the source folder is copied to the Panorama Public copy.
- Added Formula and ChemElement classes to build / add / subtract and normalize modification formulas
- Added test coverage

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/proteomexchange/UnimodModification.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/proteomexchange/UnimodModification.java
@@ -17,33 +17,33 @@ package org.labkey.panoramapublic.proteomexchange;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
+import org.labkey.api.util.Link;
 
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static org.labkey.panoramapublic.proteomexchange.UnimodParser.*;
+import static org.labkey.panoramapublic.proteomexchange.UnimodParser.Position;
+import static org.labkey.panoramapublic.proteomexchange.UnimodParser.Specificity;
+import static org.labkey.panoramapublic.proteomexchange.UnimodParser.TermSpecificity;
+import static org.labkey.panoramapublic.proteomexchange.UnimodParser.Terminus;
 
 public class UnimodModification
 {
     private final int _id;
     private final String _name;
-    private final String _normFormula;
+    private final Formula _formula;
     private final Set<Specificity> _modSites;
     private TermSpecificity _nTerm;
     private TermSpecificity _cTerm;
     private boolean _isIsotopic;
 
-    public UnimodModification(int id, String name, String normalizedFormula)
+    public UnimodModification(int id, String name, Formula formula)
     {
         _id = id;
         _name = name;
-        _normFormula = normalizedFormula;
+        _formula = formula;
         _modSites = new HashSet<>();
     }
 
@@ -57,9 +57,14 @@ public class UnimodModification
         return _name;
     }
 
+    public Formula getFormula()
+    {
+        return _formula;
+    }
+
     public String getNormalizedFormula()
     {
-        return _normFormula;
+        return _formula.getFormula();
     }
 
     public void setNterm(@NotNull Position position)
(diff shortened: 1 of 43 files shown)


panoramapublic/src/org/labkey/panoramapublic/proteomexchange/UnimodModification.java
```diff
@@ -155,147 +160,93 @@ public boolean matches(String normFormula, @NotNull Set<Specificity> sites, Term
...
-    public String getModSites()
+    public static Link getLink(int unimodId)
     {
-        if(_modSites.size() > 0)
+        return getLink(unimodId, false);
+    }
+
+    public static Link getLink(int unimodId, boolean clearClasses)
+    {
+        var link = new Link.LinkBuilder("UNIMOD:" + unimodId)
```
labkey-jeckels, line 243: For security reasons, please include rel="noopener noreferrer" for links that use target="_blank"

https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable

panoramapublic/src/org/labkey/panoramapublic/query/modification/NormalizedFormulaDisplayColumnFactory.java
```diff
@@ -0,0 +1,37 @@
...
+    {
+        return new DataColumn(colInfo)
+        {
+            @Override
+            public void renderGridCellContents(RenderContext ctx, Writer out) throws IOException
+            {
+                String formula = ctx.get(colInfo.getFieldKey(), String.class);
+                if (formula != null)
+                {
+                    out.write(Formula.normalizeFormula(formula));
```
labkey-jeckels: Perhaps this will never container special characters, but I believe this should be HTML-encoded just to be safe.

labkey-jeckels: Also, this won't affect Excel or TSV exports, or API calls against this column. Do you care about those scenarios? If so, you'll want to hook into getValue().

panoramapublic/src/org/labkey/panoramapublic/view/combinationModInfo.jsp
```diff
@@ -0,0 +1,327 @@
...
+            if (!el.hasCls(alertCls)) el.addCls(alertCls);
+
+            el.removeCls(formulaBalanced ? alertWarnCls : alertInfoCls);
+            el.addCls(formulaBalanced ? alertInfoCls : alertWarnCls);
+            el.update(html);
+        }
+    }
+
+    function span(str) {
+        return '<span style="margin: 0 10px 0 10px;font-weight:bold;">' + str + '</span>';
```
labkey-jeckels: Perhaps also known to be safe, but probably also wise to HTML encode str here

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/proteomexchange/UnimodModification.java
vagisha: OK, thanks.  I have updated pxValidationStatus.jsp as well.

panoramapublic/src/org/labkey/panoramapublic/query/modification/NormalizedFormulaDisplayColumnFactory.java
vagisha: Thanks for the reminder!  Yes, I would like to be able to see the normalized formula in Excel / TSV exports.  Overriding `getValue()`, `getDisplayValue()` and `getFormattedHtml()` now.

panoramapublic/src/org/labkey/panoramapublic/view/combinationModInfo.jsp
vagisha: Done.

PR #223, opened 2022-07-19
Issue 45903: Report Skyline document downloads to Google Analytics 4

labkey-jeckels:
#### Rationale
We have a couple of places where our generated HTML is submitting special events to Google Analytics. We're switching to GA4 as Google has deprecated the old style.

#### Changes
* Co-submit to Universal Analytics and GA4


labkey-jeckels:
@vagisha I'm not set up to test these locally but I did test the Skyline document download codepath in the TargetedMS module so I'm confident in the approach.

vagisha (approved):

PR #237, opened 2022-08-17
Show the "Make Public" button in the Panorama Public copy of a dataset

vagisha:
#### Rationale
Folders submitted to Panorama Public have a "Make Public" button that is displayed after the folder is copied to Panorama Public. Clicking the button makes the copied folder, in the Panorama Public project, public and allows the user to add publication information.  It would be useful to have this button in the Panorama Public copy, and give the submitter permissions to make their data public from the folder in Panorama Public.

#### Related Pull Requests
https://github.com/LabKey/MacCossLabModules/pull/135

#### Changes
- The PanoramaPublicSubmitterRole is assigned to the submitter and lab head users when a folder is copied to Panorama Public. The "Make Public" button is displayed in the Panorama Public copy if the logged in user has PanoramaPublicSubmitterPermission
- Added test
- Bug fixes
  - Fixed link to user details page in the data validation summary panel 
  - Deleting the previous copy of a folder did not work if the folder had sub-folders

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/security/PanoramaPublicSubmitterRole.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/security/PanoramaPublicSubmitterRole.java
@@ -1,9 +1,12 @@
 package org.labkey.panoramapublic.security;
 
+import org.labkey.api.data.Container;
 import org.labkey.api.security.SecurableResource;
 import org.labkey.api.security.SecurityPolicy;
 import org.labkey.api.security.permissions.ReadPermission;
 import org.labkey.api.security.roles.AbstractRole;
+import org.labkey.panoramapublic.query.ExperimentAnnotationsManager;
+import org.labkey.panoramapublic.query.JournalManager;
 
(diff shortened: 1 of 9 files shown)


labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/security/PanoramaPublicSubmitterRole.java
```diff
@@ -19,6 +22,14 @@ public PanoramaPublicSubmitterRole()
     @Override
     public boolean isApplicable(SecurityPolicy policy, SecurableResource resource)
     {
-        return false; // Do not show in the permission UI
+        if (super.isApplicable(policy, resource))
+        {
+            // Show the role on the permissions page of subfolders of the Panorama Public project that have an experiment.
+            Container project = ((Container)resource).getProject();
```
labkey-jeckels, line 28: You should only get here if the resource is a Container (as verified by the superclass) - maybe worth a comment to that effect.

labkey-jeckels (approved):

PR #247, opened 2022-09-15
Automation test for Panorama Web Public Search Improvements

labkey-sweta:
#### Rationale
Automation test for Panorama Web Public Search Improvements

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
+++ b/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
@@ -399,6 +399,7 @@
                                                 fieldLabel: 'Exact Matches Only',
                                                 name: exactProteinMatchesItemId,
                                                 itemId: exactProteinMatchesItemId,
+                                                id: exactProteinMatchesItemId,
                                                 input: true,
                                                 // labelAlign: 'right',
                                                 boxLabelAlign: 'after',
@@ -465,6 +466,7 @@
                                                 fieldLabel: 'Exact Matches Only',
                                                 name: exactPeptideMatchesItemId,
                                                 itemId: exactPeptideMatchesItemId,
+                                                id: exactPeptideMatchesItemId,
                                                 input: true,
                                                 labelWidth: 130,
                                                 listeners: {
--- a/panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
+++ b/panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
@@ -7,9 +7,6 @@ import org.labkey.test.components.html.Input;
 import org.labkey.test.util.DataRegionTable;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
-import org.openqa.selenium.support.ui.ExpectedConditions;
-
-import java.util.Optional;
 
 public class PanoramaPublicSearchWebPart extends BodyWebPart<PanoramaPublicSearchWebPart.ElementCache>
 {
@@ -124,13 +121,7 @@ public class PanoramaPublicSearchWebPart extends BodyWebPart<PanoramaPublicSearc
 
     public DataRegionTable search()
     {
-        Optional<DataRegionTable> oldResults = new DataRegionTable.DataRegionFinder(getDriver()).findOptional(this);
         elementCache().search.click();
-        if (oldResults.isPresent())
-        {
-            getWrapper().shortWait().until(ExpectedConditions
-                    .stalenessOf(oldResults.get().getComponentElement()));
-        }
         return new DataRegionTable.DataRegionFinder(getDriver()).waitFor(this);
     }
 
--- a/panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
+++ b/panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
@@ -1,0 +1,25 @@
+package org.labkey.test.components.panoramapublic;
+
+import org.labkey.test.components.BodyWebPart;
+import org.openqa.selenium.WebDriver;
+
+public class PanoramaPublicSearchWebPart extends BodyWebPart<PanoramaPublicSearchWebPart.ElementCache>
+{
+    private static String title = "Panorama Public Search";
+
+    public PanoramaPublicSearchWebPart(WebDriver driver, String title)
+    {
+        super(driver, title);
+    }
+
+    @Override
+    protected PanoramaPublicSearchWebPart.ElementCache newElementCache()
+    {
+        return new PanoramaPublicSearchWebPart.ElementCache();
+    }
+
+    protected class ElementCache extends BodyWebPart.ElementCache
+    {
+
+    }
+}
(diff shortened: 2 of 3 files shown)


panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
```diff
@@ -399,6 +399,7 @@
                                                 fieldLabel: 'Exact Matches Only',
                                                 name: exactProteinMatchesItemId,
                                                 itemId: exactProteinMatchesItemId,
+                                                id: exactProteinMatchesItemId,
```
labkey-tchad, line 402: Shouldn't use hard-coded element IDs here.
WebParts are not guaranteed to be unique but HTML IDs should be.

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
```diff
@@ -0,0 +1,154 @@
...
+    {
+        elementCache().proteinTab.click();
+        return this;
+    }
+
+    public PanoramaPublicSearchWebPart gotoPeptideSearch()
+    {
+        elementCache().peptideTab.click();
+        return this;
+    }
```
labkey-tchad, line 120: Is there something to wait for when switching to protein or peptide search? Similar to `gotoExperimentSearch` waiting for the "Authors" to be visible.

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
```diff
@@ -0,0 +1,154 @@
...
+    public DataRegionTable search()
+    {
+        elementCache().search.click();
+        return new DataRegionTable.DataRegionFinder(getWrapper().getDriver()).find(this);
+    }
+
+    public DataRegionTable refreshedTable()
+    {
+        return new DataRegionTable.DataRegionFinder(getWrapper().getDriver()).find(this);
+    }
```
labkey-tchad: Should wait for search results grid to update. Assuming the grid doesn't appear until you search once; probably something like:
```suggestion
    public DataRegionTable search()
    {
    
        Optional<DataRegionTable> resultsGrid = new DataRegionTable.DataRegionFinder(getWrapper().getDriver()).findOptional(this);
        if (resultsGrid.isPresent())
        {
            resultsGrid.get().doAndWaitForUpdate(() -> elementCache().search.click());
            return resultsGrid.get();
        }
        else
        {
            elementCache().search.click();
            return findResultsGrid();
        }
    }

    public DataRegionTable findResultsGrid()
    {
        return new DataRegionTable.DataRegionFinder(getWrapper().getDriver()).waitFor(this);
    }
```

panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaWebPublicSearchTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Test
+    public void testExperimentalSearch()
+    {
+        goToProjectHome();
+        PanoramaPublicSearchWebPart panoramaPublicSearch = new PanoramaPublicSearchWebPart(getDriver(), "Panorama Public Search");
+        DataRegionTable table = panoramaPublicSearch.setAuthors(AUTHOR_FIRST_NAME).search();
+        CustomizeView customizeView = table.openCustomizeGrid();
+        customizeView.addColumn("Authors");
+        customizeView.clickViewGrid();
+        table = panoramaPublicSearch.refreshedTable();
```
labkey-tchad: `DataRegionTable.applyCustomView` will wait for the data region to update. (`clickViewGrid` should probably be deprecated at some point since it does not)
```suggestion
        customizeView.applyCustomView(0);
```

In general, I don't think the `refreshedTable` method should be necessary as long as you use methods that wait for the data region to update.

labkey-tchad (changes requested):

panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
labkey-sweta: I did that change based on Ankur's suggestion, i will defer this to him @ankurjuneja

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
labkey-sweta: I had added initially, but it was obsolete so removed it.

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
labkey-sweta: Done

panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaWebPublicSearchTest.java
labkey-sweta: Done

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
labkey-sweta: After doing the above change, test started failing waiting for cssSignal for doAndWaitForUpdate

panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
ankurjuneja: @labkey-tchad this is an ext js element inside WebPart. I also found that we can't use the id for LABKEY.Webpart but I thought we can use ids for ext js items.

labkey-tchad: @ankurjuneja I understand that but if you had two webparts, they would each have an ext js element with `id = "exactPeptideMatches-inputEl"`. Ext might not actually care about the ID but I've seen React behave strangely when two inputs have the same id.
Regardless, HTML ids are supposed to be unique.

panoramapublic/test/src/org/labkey/test/components/panoramapublic/PanoramaPublicSearchWebPart.java
```diff
@@ -0,0 +1,149 @@
...
+    {
+        elementCache().peptideTab.click();
+        return this;
+    }
+
+    public DataRegionTable search()
+    {
+        elementCache().search.click();
+        return new DataRegionTable.DataRegionFinder(getWrapper().getDriver()).waitFor(this);
+    }
```
labkey-tchad: Looking at the failure screenshot. It seems like searching creates a replacement data region instead of updating the existing one. Try waiting for the previous one to go stale when searching.
```suggestion
    public DataRegionTable search()
    {
        Optional<DataRegionTable> oldResults = new DataRegionTable.DataRegionFinder(getDriver()).findOptional(this)
        elementCache().search.click();
        if (oldResults.isPresent())
        {
            getWrapper().shortWait().until(ExpectedConditions
                    .stalenessOf(oldResults.get().getComponentElement()));
        }
        return new DataRegionTable.DataRegionFinder(getDriver()).waitFor(this);
    }
```

labkey-tchad (approved):
Well. That fix didn't work either.
It's odd that it worked for the protein and peptide search tests but the experiment search is failing.
Oh well 🤷

ankurjuneja (approved):

PR #252, opened 2022-09-23
Issue 46374: Discrepancy in testresults schema

labkey-jeckels:
#### Rationale
This schema is reporting mismatch on various dev machines. Everything matches when you're installing from scratch and on skyline.ms.

#### Changes
* Ensure table ends up in the state intended from testresults-13.29-13.30.sql


labkey-adam (approved):
Tested locally as well. Feel free to resolve issue to me.

PR #278, opened 2022-11-18
Fix column name mismatch

vagisha:
Column names in the lincsmetadata table on PanoramaWeb and PanoramaWeb-DR do not match the metadata in lincs.xml. In the database the table has a 'file' column whereas in the metadata the column name is 'FileName'.  I do not know when or why the column was renamed on PanoramaWeb.  But it showed up as an error when PanoramaWeb-DR was upgraded to 22.11. 

The file/filename column contains the name of the .sky.zip file, and isn't actually used in the code. This table was used by the LINCS project members to enter additional, plate-level metadata for uploaded Skyline documents. 

Changes:
- Change the column name in lincs.xml
- It is unlikely that the LINCS module in deployed on any other servers. But if it is, rename the column.

### DIFF
--- a/lincs/resources/schemas/dbscripts/postgresql/lincs-22.000-22.001.sql
+++ b/lincs/resources/schemas/dbscripts/postgresql/lincs-22.000-22.001.sql
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2019 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
As you may know, I added validation that the DB schema matches what's in the schema XML file as part of 22.11. This has likely been a lurking issue for a while that's just now being reported.

vagisha:
> As you may know, I added validation that the DB schema matches what's in the schema XML file as part of 22.11. This has likely been a lurking issue for a while that's just now being reported.

Ah, I thought that validation was already part of the deployment process, so I was a bit surprised to see the error.  Thanks for adding it!

labkey-jeckels:
@vagisha TeamCity has long run automated tests that compare the DB schema and the XML file. It's new that we run this on production servers. It's intended to catch cases like this, where the current bootstrap scripts match the XML but somewhere along the way, the upgrade scripts didn't execute as expected.

PR #306, opened 2023-02-06
Issue 42270: Promo figures for Panorama Public datasets

vagisha:
#### Rationale
We want to request submitters to give us a short text and an image describing their Panorama Public dataset when they make it public. This will be included in the slideshow on the PanoramaWeb homepage. 

#### Changes
- After making their data public, submitters will see a button they can click to provide an entry (description + image) for the slideshow catalog.
- Enable cropping the selected image to required aspect ratio in the catalog entry form before uploading to the server.
- A CatalogEntryWebPart is added to the experiment details page.  This webpart, visible only to folder admins and Panorama Public submitters, will display the entry.  Or, if an entry does not exist for a public dataset, the webpart will display a button to add an entry.
- A Panorama Public admin can review the entry and approve or reject in the CatalogEntryWebPart or in the CatalogEntryTableInfo query.
- The Panorama Public admin console has a link for "Panorama Public Catalog Settings" that can be used to configure the max image file size, image dimensions, character limit for the description, and the maximum entries to display in the slideshow.
- GetCatalogApiAction action can be used for getting a list of approved entries in JSON format. A test page (viewSlideshow.jsp) has been added as an example use case.


panoramapublic/src/org/labkey/panoramapublic/query/CatalogEntryTableInfo.java
```diff
@@ -0,0 +1,196 @@
...
+                    viewButton.appendTo(out);
+                    return;
+                }
+                super.renderGridCellContents(ctx, out);
+            }
+            @Override
+            public void addQueryFieldKeys(Set<FieldKey> keys)
+            {
+                super.addQueryFieldKeys(keys);
+                keys.add(FieldKey.fromParts("Id"));
```
labkey-jeckels: I'm not seeing where `Id` is being used. If it's needed, it's best practice to prefix it with the current column's FieldKey's parent too, as that will make it more likely to work if the table has been included in a custom query, as a lookup, etc.

panoramapublic/src/org/labkey/panoramapublic/query/CatalogEntryTableInfo.java
```diff
@@ -0,0 +1,196 @@
...
+                    return;
+                }
+                super.renderGridCellContents(ctx, out);
+            }
+
+            @Override
+            public void addQueryFieldKeys(Set<FieldKey> keys)
+            {
+                super.addQueryFieldKeys(keys);
+                keys.add(FieldKey.fromParts("ShortUrl"));
```
labkey-jeckels: Similar advice here - include the current column's FieldKey parent both here and when you pull it during `renderGridCellContents()`

panoramapublic/src/org/labkey/panoramapublic/view/publish/catalogEntryForm.jsp
```diff
@@ -0,0 +1,371 @@
...
+                    <input id="imageFileInput" type="file" size="50" style="border: none; background-color: transparent;" accept="image/png,image/jpeg" />
+                    <input id="modifiedImage" name="imageFile" type="hidden"/>
+                    <input id="imageFileName" name="imageFileName" type="hidden"/>
+                    <div id="preview" style="margin-top:10px; padding:10px;"></div>
+                </td>
+            </tr>
+        </table>
+        <br>
+        <%=button("Submit").submit(true).disableOnClick(true)%>
+        &nbsp
```
labkey-jeckels:
```suggestion
        &nbsp;
```

labkey-jeckels (approved):

PR #309, opened 2023-02-08
Assign PanoramaPublicSubmitterRole to data submitters of older datasets

vagisha:
#### Rationale
The PanoramaPublicSubmitterRole allows submitters and lab heads to make their data public in the Panorama Public project, and add a slideshow entry for a dataset that is already public.  This role is assigned to the submitter and lab head when data is copied to Panorama Public (https://github.com/LabKey/MacCossLabModules/pull/237).  We need to assign the role to submitters of older datasets that were copied before this role was added.

#### Related Pull Requests
- https://github.com/LabKey/MacCossLabModules/pull/237
- https://github.com/LabKey/MacCossLabModules/pull/306

#### Changes
- Added a pipeline job that will iterate through the folders in a project, and update permissions for submitter and lab head in folders that have a valid experiment.  This job can be run by site admins from the Panorama Public Admin Console.
- Renamed JournalGroupsAdminViewAction to PanoramaPublicAdminViewAction.
- Removed AddPanoramaPublicModuleAction. This was used to add the panoramapublic module to older targetedms folders, and is no longer needed. The panoramapublic module gets added when a TargetedMSFolder is created (TargetedMSFolderTypeListener.folderCreated implementation in PanoramaPublicListener).


panoramapublic/src/org/labkey/panoramapublic/pipeline/AssignSubmitterPermissionJob.java
```diff
@@ -0,0 +1,135 @@
...
+            {
+                logger.info(String.format("'%s', %s: %s - %s", container.getPath(), userType, user.getEmail(), "assigning"));
+                MutableSecurityPolicy newPolicy = new MutableSecurityPolicy(container, container.getPolicy());
+                newPolicy.addRoleAssignment(user, PanoramaPublicSubmitterRole.class, false);
+                SecurityPolicyManager.savePolicy(newPolicy);
+                return true;
+            }
+            else
+            {
+                logger.info(String.format("'%s', %s: %s - %s", container.getPath(), userType, user.getEmail(), "would assign (dry run)"));
```
labkey-jeckels, line 114: From my reading of the code, it should be iterating through the containers depth-first. That's good, it means that the role assignment may be added at the project level and then cascade down any children that inherit permissions.

However, I think the dry run will log that it's adding it to all children even though it won't need to. If you don't care, that's fine, but you could track the resource ID of the policy used for a container and not log (or log differently) when you revisit one.

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/pipeline/AssignSubmitterPermissionJob.java
vagisha: It should be OK since the role will only be assigned in containers that have a TargetedMSExperiment.  All other containers will be logged as not having a valid experiment.

PR #320, opened 2023-03-02
Promo figures for Panorama Public datasets - Updates

vagisha:
#### Rationale
Some updates to the UI for collecting Panorama Public data catalog entries

#### Related Pull Requests
- https://github.com/LabKey/MacCossLabModules/pull/306

#### Changes
- In the success view after data is made public, display a slideshow preview to encourage submitter to provide a catalog entry.
- Added a catalog entry icon to the Targeted MS Experiment webpart. This links to either the action to add a new catalog entry, or view the existing one. It is displayed only in the most recent version of the data in the case that there are multiple copies of a dataset (no such datasets yet on Panorama Public).
- Updated text in the catalog entry form.
- Post message to support thread when a catalog entry is added.
- Added "View Slideshow" button in the Catalog Entry table toolbar. Visible only to site admins. This can used to preview the pending entries, for example.
  - Updated viewSlideshow.jsp - added form elements to choose the catalog entry type (all, approved, rejected, pending) and the number of entries to display.
- Moved JS and CSS from viewSlideshow.jsp to source files so that they can be reused.
- Added test.


panoramapublic/src/org/labkey/panoramapublic/query/CatalogEntryManager.java
```diff
@@ -144,9 +147,10 @@ public static void moveEntry(@NotNull ExperimentAnnotations previousCopy, @NotNu
         }
     }
 
-    public static List<CatalogEntry> getApprovedEntries(int entryCount)
+    public static List<CatalogEntry> getEntries(@NotNull CatalogEntryType entryType, int entryCount)
```
labkey-jeckels, line 150: Does entryCount just cap the number of rows it should return? If so, I think you could call `setMaxRows(entryCount)` on the `TableSelector`

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsManager.java
```diff
@@ -543,6 +543,20 @@ public static boolean hasProteomicData(ExperimentAnnotations experimentAnnotatio
...
+    /**
+     * @param experimentAnnotations
+     * @return true if the given experiment is a journal copy, and is the most recent version.
+     */
+    public static boolean isCurrentVersion(ExperimentAnnotations experimentAnnotations)
+    {
+        if (experimentAnnotations != null && experimentAnnotations.isJournalCopy())
+        {
+            Integer maxVersion = ExperimentAnnotationsManager.getMaxVersionForExperiment(experimentAnnotations.getSourceExperimentId());
+            return experimentAnnotations.getDataVersion().equals(maxVersion);
```
labkey-jeckels, line 555: Is `getDataVersion()` guaranteed to be non-null? If so, should it return an `int` instead of `Integer`?

panoramapublic/src/org/labkey/panoramapublic/view/expannotations/experimentDetails.jsp
```diff
@@ -191,6 +212,19 @@
...
      display: block;
  }
+ a.catalogIcon {
+     display: inline-block;
+     height: 20px;
+     width: 25px;
+ }
+ a.catalogIconGreen
+ {
+     background: url("../../PanoramaPublic/images/slideshow-icon-green.png") no-repeat bottom 0 right 0;
```
labkey-jeckels: Will this always be the right path? I worry about the `../../` part. Might be better to make it something like:

```suggestion
     background: url("<%= h(contextPath) %>/PanoramaPublic/images/slideshow-icon-green.png") no-repeat bottom 0 right 0;
```

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsManager.java
vagisha: It is non-null only for experiments that are copies in the Panorama Public project (isJournalCopy()).

panoramapublic/src/org/labkey/panoramapublic/query/CatalogEntryManager.java
vagisha: Yes, this could be simplified with `setMaxRows(entryCount)` and `setOffset(offset)`.  I will do that in the next iteration.

panoramapublic/src/org/labkey/panoramapublic/view/expannotations/experimentDetails.jsp
vagisha: Done.

PR #323, opened 2023-03-07
Remove 'Clear Variables' button on Peptide and Protein search results grid

labkey-bpatel:
#### Rationale
Clicking on 'Clear Variables' button on Peptide and Protein search results grid is showing underlying parameterized query view.

#### Changes
* Hide 'clear variables' button on Peptide and Protein search results grid

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
+++ b/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
@@ -349,32 +349,21 @@
             if (activeTab === expSearchPanelItemId) {
                 parseUrlQueryParams()
             }
-            if (activeTab === proteinSearchPanelItemId) {
-
-                if (context[proteinNameItemId] === undefined) {
-                    context[proteinNameItemId] = '';
-                }
-                proteinParameters[proteinNameItemId] = context[proteinNameItemId];
+            if (context[proteinNameItemId] || context[proteinNameItemId] === '') {
+                proteinParameters[proteinNameItemId] =  context[proteinNameItemId];
                 document.getElementById(proteinNameItemId).value = context[proteinNameItemId];
-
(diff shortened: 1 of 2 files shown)


labkey-jeckels (changes requested):
This feels like it's going to be fragile in the face of changes to the HTML or CSS for the grid.

Could we use the `showFilterDescription` property to hide the whole header from the grid? Or are we relying on that header to show what the user searched for? I'm generally OK with relying on the search from section above for that context.

[showFilterDescription](https://www.labkey.org/download/clientapi_docs/javascript-api/symbols/LABKEY.QueryWebPart.html#constructor)

Whatever approach, perhaps we could add test coverage to make sure it's not showing?

labkey-bpatel:
> This feels like it's going to be fragile in the face of changes to the HTML or CSS for the grid.
> 
> Could we use the `showFilterDescription` property to hide the whole header from the grid? Or are we relying on that header to show what the user searched for? I'm generally OK with relying on the search from section above for that context.
> 
> [showFilterDescription](https://www.labkey.org/download/clientapi_docs/javascript-api/symbols/LABKEY.QueryWebPart.html#constructor)
> 
> Whatever approach, perhaps we could add test coverage to make sure it's not showing?

We are relying on `showFilterDescription` to show what user searched for. Test is already updated to test that its not visible.

labkey-jeckels (approved):
I'm sorry, I don't know how I missed the test updates in my initial review.

They seem to be checking the same potentially fragile CSS class though. Can we also make sure that the "Clear Variables" text from that button isn't showing?

labkey-bpatel:
> I'm sorry, I don't know how I missed the test updates in my initial review.
> 
> They seem to be checking the same potentially fragile CSS class though. Can we also make sure that the "Clear Variables" text from that button isn't showing?

Made a change so that 'Clear Variables' button gets removed instead of hidden, hence the text from that button is not present any longer. Test is updated as well.

PR #325, opened 2023-03-08
Signup module updates

vagisha:
#### Changes
- Links in the the labkeyuserid column were broken. Added UserIdQueryForeignKey to the column.
- Changed "containerId" parameter name to "folderId" since containerId is not allowed (HasAllowBindParameter.disallowed). AddPropertyAction and RemovePropertyAction were broken as a result because the value of containerId was not getting passed though. 
- ConfirmAction is now a FormViewAction. User is presented with a form where they provide a valid password. A new labkey user is created and assigned the password.
   - Code was copied from LoginController.attemptSetPassword(). Should it be moved to SecurityManager or AuthenticationManager?
   - Consider moving DbLoginManager from platform.core.main to platform.api.main so that we can get the PasswordRule on the server.

### DIFF
--- a/signup/resources/schemas/signup.xml
+++ b/signup/resources/schemas/signup.xml
@@ -28,14 +28,26 @@
             <column columnName="organization"/>
             <column columnName="key"/>
             <column columnName="container"/>
-            <column columnName="labkeyuserid"/>
+            <column columnName="labkeyuserid">
+                <fk>
+                    <fkColumnName>UserId</fkColumnName>
+                    <fkDbSchema>core</fkDbSchema>
+                    <fkTable>UsersData</fkTable>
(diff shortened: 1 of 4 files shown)


labkey-jeckels (approved):
This approach looks OK to me. I probably wouldn't move all of DbLoginManager (it would drag at least DbLoginConfiguration and maybe more) from Core to API, but since PasswordRule is already in API it should be pretty easy to expose a way to get the current value on a Service of some sort. Up to you if it's worth the effort since this the signup module is only deployed to servers that you manage.

vagisha:
Thanks for the review.  I will merge this for now and look at exposing the current value of PasswordRule later.

PR #333, opened 2023-03-26
Panorama Public updates and bug fixes

vagisha:
#### Changes
- Add submitters to the "Panorama Public Submitters" group. Add reviewer to "Reviewers" group. These groups exist on panoramaweb.org.
- Replaced "Access URL / Link" with "Permanent Link".
- Added text about license (e.g. CC-BY) in the confirm view when data is made public.
- Fixed container links and "Make Public" link in notifications.  These have to be absolute URLs.
- Added DOI in the message body of the data-copied notification.
- Display DOI as a complete link as recommended by the DataCite DOI Display Guidelines (https://support.datacite.org/docs/datacite-doi-display-guidelines).

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
@@ -38,6 +38,7 @@ import org.labkey.api.portal.ProjectUrls;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.security.Group;
+import org.labkey.api.security.InvalidGroupMembershipException;
 import org.labkey.api.security.MutableSecurityPolicy;
 import org.labkey.api.security.SecurityManager;
 import org.labkey.api.security.SecurityPolicy;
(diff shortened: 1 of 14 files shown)


panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
```diff
@@ -524,11 +536,41 @@ private void updatePermissions(User formSubmitter, ExperimentAnnotations targetE
...
+            // This group already exists in the Panorama Public project on panoramaweb.org.
+            // CONSIDER: Make this configurable through the Panorama Public admin console.
+            addToGroup(user, "Panorama Public Submitters", project, log);
+        });
+    }
+
+    private void addToGroup(User user, String groupName, Container project, Logger log)
+    {
+        Integer groupId = SecurityManager.getGroupId(project, groupName, false);
+        if (groupId != null)
```
labkey-jeckels: Would it be useful to log if we can't find the group? Does the test create it today?

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
vagisha: Done.

labkey-jeckels (approved):

PR #344, opened 2023-04-07
Issue 47144: Handle large files on Panorama Public via Symlinks

labkey-martyp:
#### Rationale
Move instead of copying files to Panorama Public.

https://www.labkey.org/Panorama%20Partners/Feature%20Requests/issues-details.view?issueId=47144

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4280

#### Changes
* Add PanoramaPublicFileImporter to handle file moves instead of copies.
* Add PanoramaPublicMetadataImporter to handle run metadata on copy
* Add PanoramaPublicSymlinkManager to handle symlinks on copies and other events
* Add symlink handling for file and container listeners
* Update and add new tests

PR #361, opened 2023-07-06
Add filesystem audit events for Panorama Public symlinks

vagisha:
#### Rationale
We would like to add filesystem audit events: 
- When a file is moved from the submitted folder to the copy on Panorama Public, and a symlink is created in the submitted folder.
- When the target of a symlink in the submitted folder is moved from the previous copy on Panorama Public to a new copy
  - add an event in the previous copy that the file has been moved, and replaced with a symlink
  - add an event in the submitted folder that the symlink target has been updated. 
- When a Panorama Public copy is deleted, add events in the submitted folder for each symlink that is replaced with the target file from the Panorama Public copy.

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/344

#### Changes
- added filesystem events
- added additional logging
- removed unused boolean parameter createSourceSymLinks in PanoramaPublicSymlinkManager.moveAndSymLinkDirectory()

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicFileImporter.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicFileImporter.java
@@ -85,8 +85,15 @@ public class PanoramaPublicFileImporter implements FolderImporter
                 Files.createDirectories(targetFiles.toPath());
             }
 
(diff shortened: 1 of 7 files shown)


labkey-martyp (approved):
Sorry @vagisha I've been out on vacation. I did not test this but I don't see any issues in the code. Let us know if you need any testing.

labkey-jeckels (approved):

PR #367, opened 2023-07-21
Panorama Public updates and fixes

vagisha:
#### Changes
- Include text and link to add a catalog entry in the Panorama Public notification message.
- Assign PXD accession and DOI at the end of CopyExperimentFinalTask so that we don't create these identifiers again in case the task has to be rerun due to a previous error.
- Fix NPE in confirmPublish.jsp if data license is null.
- Fix NPE if the dataVersion of an experiment is null. This can happen if there is an error in the data copy pipeline job.
- Log names of users that were assigned Reader role in the Panorama Public copy. Always include the form submitter and users assigned as data submitter and lab head. 
- Added more file audit events. Set directory and file on audit event.  Fixed container path in log messages.
- When a Panorama Public container is deleted look at all containers and sub-containers that may have a symlink target in the container being deleted.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/view/publish/confirmPublish.jsp
+++ b/panoramapublic/src/org/labkey/panoramapublic/view/publish/confirmPublish.jsp
@@ -4,7 +4,6 @@
 <%@ page import="org.labkey.api.view.template.ClientDependencies" %>
 <%@ page import="org.labkey.panoramapublic.PanoramaPublicController" %>
 <%@ page import="org.labkey.api.portal.ProjectUrls" %>
-<%@ page import="org.labkey.api.util.HtmlString" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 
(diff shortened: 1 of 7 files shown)


panoramapublic/src/org/labkey/panoramapublic/view/publish/confirmPublish.jsp
```diff
@@ -32,11 +33,8 @@
         var items = [];
         if (<%=!bean.isPublic()%>) {
             let html = 'Data at ' + <%=qh(bean.getAccessUrl())%> + ' will be made public.';
-            if (<%= bean.getLicense() != null %>) {
-                html += ' It will be available under the ' + <%=qh(bean.getLicense().getDisplayName())%> + ' license.';
-            }
+            html += <%= bean.getLicense() != null ? qh("It will be available under the " + bean.getLicense().getDisplayName() + ".") : HtmlString.EMPTY_STRING %>;
```
labkey-jeckels: Maybe I'm misreading the code, but if the license is null, I think this will render as:

`html += ;`

which isn't valid JS.

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/view/publish/confirmPublish.jsp
vagisha: Indeed.  It has been fixed. Thanks for catching.

labkey-jeckels (approved):
The updates look good.

PR #373, opened 2023-08-16
Don't kill folder import

labkey-matthewb:
#### Rationale
I'm not exactly sure what the folder import flow is, but this seems to kill lots of folder imports (bad for passing daily suites on a dev machine).

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicFileImporter.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicFileImporter.java
@@ -61,12 +61,6 @@ public class PanoramaPublicFileImporter implements FolderImporter
             return;
         }
 
-        if (null == job)
-        {
-            log.error("Pipeline job not found.");
-            return;
-        }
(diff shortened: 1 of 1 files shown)


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicFileImporter.java
```diff
@@ -61,12 +61,6 @@ public void process(@Nullable PipelineJob job, FolderImportContext ctx, VirtualF
             return;
         }
 
-        if (null == job)
-        {
-            log.error("Pipeline job not found.");
-            return;
-        }
-
```
labkey-martyp, line 69: I see. This check was for a specific panorama work flow to copy a folder. Looks like for some folder imports, like container from template, the job is null. In that case the folder importer will roll through all these importers with a null job, so that would be an expected state. So removing this makes sense.

labkey-martyp (approved):

PR #375, opened 2023-08-28
Issue 48499: Use preferred SecurityPolicyManager.savePolicy() variant

labkey-jeckels:
#### Rationale
We're uneven in terms of the validation and auditing we do for saving SecurityPolicies and related updates, and want to be more consistent.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4704

#### Changes
* Let modules register ContainerSecurableResourceProviders instead of having Container know about them all
* Remove the `savePolicy` and `createContainer` methods that don't take a user, check permissions, or log for audit purposes
* Introduce `User.getAdminServiceUser()` for `sudo` like scenarios or when we're doing an operation not initiated by a user, like bootstrapping the server
* Update callers

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
@@ -544,7 +544,7 @@ public class CopyExperimentFinalTask extends PipelineJob.Task<CopyExperimentFina
         assignPanoramaPublicSubmitterRole(newPolicy, log, targetExperiment.getSubmitterUser(), targetExperiment.getLabHeadUser(),
                 formSubmitter); // User that submitted the form. Can be different from the user selected as the data submitter
 
-        SecurityPolicyManager.savePolicy(newPolicy);
+        SecurityPolicyManager.savePolicy(newPolicy, User.getAdminServiceUser());
 
         addToSubmittersGroup(target.getProject(), log, targetExperiment.getSubmitterUser(), targetExperiment.getLabHeadUser(), formSubmitter);
     }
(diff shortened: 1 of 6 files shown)


labkey-nicka (approved):

panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
```diff
@@ -544,7 +544,7 @@ private void updatePermissions(User formSubmitter, ExperimentAnnotations targetE
         assignPanoramaPublicSubmitterRole(newPolicy, log, targetExperiment.getSubmitterUser(), targetExperiment.getLabHeadUser(),
                 formSubmitter); // User that submitted the form. Can be different from the user selected as the data submitter
 
-        SecurityPolicyManager.savePolicy(newPolicy);
+        SecurityPolicyManager.savePolicy(newPolicy, pipelineJobUser);
```
vagisha, line 547: It should be OK to pass in the User that started the pipeline job instead of User.getAdminServiceUser().  I will revert the changes in my commit if tests fail due to this change.

panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
```diff
@@ -650,11 +650,11 @@ private User createReviewerAccount(String reviewerEmailPrefix, String password,
         return newUser.getUser();
     }
 
-    private void assignReader(UserPrincipal reader, Container target)
+    private void assignReader(UserPrincipal reader, Container target, User pipelineJobUser)
     {
         MutableSecurityPolicy newPolicy = new MutableSecurityPolicy(target, target.getPolicy());
         newPolicy.addRoleAssignment(reader, ReaderRole.class);
-        SecurityPolicyManager.savePolicy(newPolicy);
+        SecurityPolicyManager.savePolicy(newPolicy, pipelineJobUser);
```
vagisha, line 657: Same here. It should be OK to pass in the User that started the pipeline job instead of User.getAdminServiceUser().

vagisha (approved):

panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
labkey-jeckels: Thanks!

PR #377, opened 2023-08-29
Switch to standard Base64 decoding class

labkey-adam:
#### Rationale
Looks like an inadvertent import... or let me know if this was intentional. I suspect the Xerces impl works correctly, but seems best to use the built-in class.


vagisha (approved):
Definitely inadvertent! Thank you for fixing.

PR #384, opened 2023-12-19
Migrate HtmlView(String) callers

labkey-jeckels:
#### Rationale
We're eliminating uses of a deprecated constructor.

#### Changes
* Switch to HtmlString or explicitly calling `unsafe()`

### DIFF
--- a/lincs/src/org/labkey/lincs/LincsController.java
+++ b/lincs/src/org/labkey/lincs/LincsController.java
@@ -63,6 +63,7 @@ import org.labkey.api.security.permissions.ReadPermission;
 import org.labkey.api.targetedms.ITargetedMSRun;
 import org.labkey.api.targetedms.SkylineAnnotation;
 import org.labkey.api.targetedms.TargetedMSService;
+import org.labkey.api.util.DOM;
 import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.Link;
 import org.labkey.api.util.PageFlowUtil;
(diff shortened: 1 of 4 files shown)


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -711,16 +712,16 @@ public ModelAndView getView(JournalSupportContainerForm form, boolean reshow, Bi
...
                 for(ObjectError error: errors.getAllErrors())
                 {
                     HtmlString errHtml = createHtmlFragment(SPAN(cl("labkey-error"), error.getDefaultMessage()));
-                    errHtml.appendTo(html);
+                    html.append(errHtml);
                 }
-                BR().appendTo(html);
-                view.addView(new HtmlView(HtmlString.unsafe(html.toString())));
+                html.append(createHtml(BR()));
+                view.addView(new HtmlView(html));
```
vagisha: I think lines 717 to 724 can be replaced with `view.addView(new HtmlView(ERRORS(errors)))`

vagisha (approved):

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
labkey-jeckels: Yes! Thanks. I forgot this was a DOM.LK method.

lincs/src/org/labkey/lincs/LincsController.java
```diff
@@ -1372,13 +1373,11 @@ public ModelAndView getView(LincsPspJobForm form, BindException errors)
...
-            sb.append("<br>JSON Output:</br>");
-            sb.append("<p><pre>").append(PageFlowUtil.filter(jsonStatus)).append("</pre></p>");
-            view.addView(new HtmlView(sb.toString()));
+            view.addView(new HtmlView(
+                    DOM.P(
+                        DOM.P("Status for job: " + pspJob.getId() +", PSP job Id: " + pspJob.getPspJobId() + ", Run Id: " + pspJob.getRunId()),
+                        DOM.BR("JSON Output:"),
+                        DOM.P(DOM.PRE(jsonStatus)))));
             view.setTitle("PSP job status");
             view.setFrame(WebPartView.FrameType.PORTAL);
```
labkey-matthewb: BR() should not have children.  This may be a mistake in the DOM api.

labkey-matthewb (changes requested):
I can fix up the BR() method and look for other cases like that.

labkey-matthewb (approved):

PR #387, opened 2024-01-08
Make it possible for users to view their data on Panorama Public

vagisha:
#### Changes
- Added a "My Data" button in the "Panorama Public Experiments" button bar. The button is displayed if a user is logged in. It links to MyDataViewAction. 
image

- My Data view displays a grid filtered to the user's public and private datasets on Panorama Public. These are experiments where the user is either a "submitter" or the "lab head". 
- The "My Data" grid displays a "Catalog Entry" column that can be used to add or view catalog entries for public datasets. Catalog entries cannot be added for private datasets.

image


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -9183,6 +9184,68 @@ public ModelAndView getView(Object o, BindException errors)
...
+            view.setShowDetailsColumn(false);
+            view.setShowDeleteButton(false);
+            view.setShowUpdateColumn(false);
+            view.setShowInsertNewButton(false);
+            view.disableContainerFilterSelection();
+            return view;
+        }
+    }
+
+    public static class MyDataForm
```
labkey-jeckels: Do we need this form at all? It seems like we only support viewing data for the current user.

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
```diff
@@ -724,4 +747,47 @@ String getRenderId()
...
+                    String imageTitle = entry != null ? "View catalog entry" : "Add catalog entry";
+                    ActionURL catalogEntryLink = entry != null ? PanoramaPublicController.getViewCatalogEntryUrl(expAnnot, entry)
+                                                               : PanoramaPublicController.getAddCatalogEntryUrl(expAnnot).addReturnURL(ctx.getViewContext().getActionURL().clone());
+                    DOM.A(at(href, catalogEntryLink.getLocalURIString(), title, PageFlowUtil.filter(imageTitle)),
+                            DOM.IMG(at(src, imageUrl, height, 22, width, 22)))
+                            .appendTo(out);
+                    return;
+                }
+            }
+            HtmlString.EMPTY_STRING.appendTo(out);
```
labkey-jeckels: This codepath uses `EMPTY_STRING` and the one above uses `NBSP`. I imagine they should probably use the same placeholder, though maybe it doesn't make a difference.

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: No, we don't need a userId since we are only displaying the data for the currently logged in user!  I removed the form.

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
vagisha: Done.

PR #390, opened 2024-01-25
Create a strong password for the reviewer account

vagisha:
#### Rationale
- We need stronger reviewer passwords.  Current passwords contain only uppercase and lowercase letters.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
@@ -18,8 +18,6 @@ package org.labkey.panoramapublic.pipeline;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
-import org.junit.Assert;
-import org.junit.Test;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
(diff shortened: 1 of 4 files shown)


panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
```diff
@@ -657,9 +661,91 @@ private void assignReader(UserPrincipal reader, Container target, User pipelineJ
...
+
+            // Add more characters until we are at the desired password length
+            while (passwordChars.size() < PASSWORD_LEN)
+            {
+                addChar(allChars, passwordChars, allChars, random);
+            }
+
+            Collections.shuffle(passwordChars);
+
+            return passwordChars.stream().map(String::valueOf).collect(Collectors.joining());
```
labkey-jeckels, line 735: It's extremely unlikely, but it's possible this could choose a password that doesn't meet our entropy threshold. You could check the generated value via `PasswordRule.Strong.isValidForLogin()` before returning it.

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/pipeline/CopyExperimentFinalTask.java
vagisha: That check is already happening in the `createPassword(User user)` method (line686) of `CopyExperimentFinalTask`.

PR #393, opened 2024-02-02
Overwrite the Content-Disposition header in DownloadToolAction for Skyline compatibility

vagisha:
#### Rationale
Skyline expects the 'Content-Disposition' response header value to look like this: filename="MSstatsShiny.zip".  Otherwise, external tool installation in Skyline does not work.

#### Changes
- Use `PageFlowUtil.streamFile` instead of returning `HttpRedirectView` in `DownloadToolAction`
- Pass 'Content-Disposition' header value to `PageFlowUtil.streamFile` so that it overwrites the value already set.

### DIFF
--- a/SkylineToolsStore/src/org/labkey/skylinetoolsstore/SkylineToolsStoreController.java
+++ b/SkylineToolsStore/src/org/labkey/skylinetoolsstore/SkylineToolsStoreController.java
@@ -67,13 +67,14 @@ import org.labkey.api.util.SafeToRender;
 import org.labkey.api.util.URLHelper;
 import org.labkey.api.view.ActionURL;
 import org.labkey.api.view.HtmlView;
-import org.labkey.api.view.HttpRedirectView;
 import org.labkey.api.view.HttpView;
 import org.labkey.api.view.JspView;
 import org.labkey.api.view.NavTree;
 import org.labkey.api.view.NotFoundException;
 import org.labkey.api.view.RedirectException;
 import org.labkey.api.view.UnauthorizedException;
+import org.labkey.api.webdav.WebdavResource;
+import org.labkey.api.webdav.WebdavService;
 import org.labkey.skylinetoolsstore.model.Rating;
 import org.labkey.skylinetoolsstore.model.SkylineTool;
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

SkylineToolsStore/src/org/labkey/skylinetoolsstore/SkylineToolsStoreController.java
```diff
@@ -1101,9 +1103,27 @@ else if (toolLsid != null &&
...
+                throw new NotFoundException("Tool container not found");
+            }
+
+            org.labkey.api.util.Path path = WebdavService.getPath().append(toolContainer.getParsedPath()).append(FileContentService.FILES_LINK).append(tool.getZipName());
+            WebdavResource resource = WebdavService.get().getResolver().lookup(path);
+            if (resource == null || !resource.isFile())
+                throw new NotFoundException("Resource could not be found: " + path.toString());
+
+            // Add our own 'Content-Disposition' header so that it overwrites the one set in
+            // ResponseHelper.setContentDisposition(HttpServletResponse response, ContentDispositionType type, @NotNull String filename)
```
labkey-matthewb, line 1119: Why not use ResponseHelper.setContentDisposition()?
Is this a matter of a client not parsing the header correctly?

A comment explaining why the code isn't use the "standard" method would be nice.

labkey-matthewb (changes requested):
see question above

SkylineToolsStore/src/org/labkey/skylinetoolsstore/SkylineToolsStoreController.java
labkey-matthewb: I just caught up with the email thread..   A reference to the issue # would be appropriate.

vagisha: This was added to the the MacCoss issues list: https://www.labkey.org/MacCoss/Issue%20Tracker/issues-details.view?issueId=49580
Added a reference to the issue # in the code.

labkey-matthewb (approved):

PR #405, opened 2024-03-17
Make the "Public" column on the ExperimentAnnotations tableinfo an ExprColumn

vagisha:
#### Rationale
The "Public" column on the ExperimentAnnotations table tells us if a Panorama Public dataset is public or not.  We have been using the hasPermissions() method on the Container for this purpose. Make the column an ExprColumn so that it is filterable.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
@@ -387,6 +387,7 @@ public class ExperimentAnnotationsTableInfo extends FilteredTable<PanoramaPublic
 
     private ExprColumn getIsPublicCol()
     {
+        // Panorama Public dataset folders do not inherit permissions from the parent folder, so we don't need to worry about that case.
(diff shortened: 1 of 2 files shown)


panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
```diff
@@ -430,6 +385,19 @@ private ExprColumn getCatalogEntryCol()
         return col;
     }
 
+    private ExprColumn getIsPublicCol()
+    {
+        SQLFragment isPublicColSql = new SQLFragment(" (SELECT CASE WHEN EXISTS (SELECT 1 FROM ")
```
labkey-jeckels, line 391: This probably won't work if a container is inheriting permissions from its parent. That may not be relevant for this use case though.

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
vagisha: I should have noted that these containers do not inherit permissions from parent, so that is not a concern.

PR #411, opened 2024-03-26
Fix "compare" function in "RunDownBean.getTopLeaks".

nickshulman:
#### Rationale
The Skyline Nightly Test Results page displays an error if there are a lot of leaks reported because the "Compare" method implementation in "getTopLeaks" does not exhibit the "transitive" property. It returns "0" when two doubles differ by less than 1, which means there are cases where A == B, B==C, but A < C.


vagisha (approved):

vagisha:
You may want to target the 23.11-SNAPSHOT branch for this one.

PR #446, opened 2024-08-21
Issue 51100: SignUp link from admin console shows up to the troubleshooter but throws 403 while accessing it

vagisha:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51100
Only site admins should be able to see the "SignUp" link in the admin console


vagisha:
Replaced with https://github.com/LabKey/MacCossLabModules/pull/447

PR #447, opened 2024-08-22
Issue 51100: SignUp link from admin console shows up to the troubleshooter but throws 403 while accessing it

vagisha:
#### Rationale

- Only site admins should be able to see the "SignUp" link in the admin console (https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51100)
- Fix failure in AdminConsoleNavigationTest due to missing Admin Console navtrails in PanoramaPublicAdminViewAction


labkey-adam (approved):
Also tested locally

PR #452, opened 2024-10-23
Fix "compare" function in "RunDownBean.getTopLeaks". (#411)

labkey-jeckels:
Rationale
The Skyline Nightly Test Results page displays an error if there are a lot of leaks reported because the "Compare" method implementation in "getTopLeaks" does not return consistent results. This was because it was mistakenly comparing the wrong set of numbers (comparing number of bytes leaked in one test to handle leaks in the other test) such that Compare(a, b) was not necessarily the opposite of Compare(b, a).

Backport of https://github.com/LabKey/MacCossLabModules/pull/411


labkey-jeckels:
Plan is to wait for the upgrade to 24.7 in a couple of weeks.

PR #503, opened 2025-02-10
Skip searching for files in 'RawFiles' if the directory is missing in a folder

vagisha:
#### Rationale
_[Issue 46118: java.nio.file.NoSuchFileException from SpecLibValidator.getPath()](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46118)_ 
For spectrum library source files we look in the "RawFiles" directory of all subfolders that have documents with the library, as well as the parent experiment folder. The exception is caused if "RawFiles" no longer exits in one of the folders being searched.

#### Changes
- Don't search for files in the "RawFiles" directory if it does not exist (e.g., if it was deleted by user) in the container.
- Added test, and test files


panoramapublic/src/org/labkey/panoramapublic/proteomexchange/validator/SpecLibValidator.java
```diff
@@ -210,6 +212,16 @@ private void validateLibrarySources(List<LibSourceFile> sources, FileContentServ
...
         Set<Container> containers = getDocsWithLibrary().stream().map(dl -> dl.getRun().getContainer()).collect(Collectors.toSet());
         containers.add(expAnnotations.getContainer());
+        Set<Path> rawFilesDirPaths = new HashSet<>();
+        for (Container container: containers)
+        {
+            var rawFiesDirPath = DataValidator.getRawFilesDirPath(container, fcs);
+            // We will look in the "RawFiles" directory only if it exists.
+            if (Files.exists(rawFiesDirPath))
+            {
+                rawFilesDirPaths.add(rawFiesDirPath);
```
labkey-jeckels:
```suggestion
            var rawFilesDirPath = DataValidator.getRawFilesDirPath(container, fcs);
            // We will look in the "RawFiles" directory only if it exists.
            if (Files.exists(rawFilesDirPath))
            {
                rawFilesDirPaths.add(rawFilesDirPath);
```

labkey-jeckels (approved):

PR #514, opened 2025-02-25
Look for report TSV file required to build spectral libraries from DIA-NN results

vagisha:
#### Rationale
Building a spectral library in Skyline from DIA-NN results requires a .speclib file and a report TSV file.  The name of the speclib file is in the SpectrumSourceFiles table of the SQLite blib file.  However, the name of the report TSV is not included.  
Skyline displays this error if the TSV file is missing while building the library:
`ERROR: unable to determine DIA-NN report filename for 'report-lib.parquet.skyline.speclib': the TSV report is required to read speclib files and must be in the same directory as the speclib and share some leading characters (e.g. somedata-tsv.speclib and somedata-report.tsv)`
The report TSV should be uploaded to pass Panorama Public data validation.

#### Changes
- For Panorama Public data validation, we will look for a TSV file, with a matching prefix, in the same directory where the speclib file was uploaded.  In addition to having a matching prefix, the TSV file should also have some of the expected column headers. 
- Added tests. Test files are heavily truncated versions of files on Panorama Public.  Since the data is not yet public, the submitter's permission was obtained to use their files to create the test data files.


panoramapublic/test/src/org/labkey/test/tests/panoramapublic/PanoramaPublicValidationTest.java
```diff
@@ -266,6 +267,117 @@ public void testLibraryValidationWithSubfolders()
...
+        DataValidationPage validationPage = submitValidationJob();
+        validationPage.verifyInvalidStatus();
+        verifySpecLibSourceFiles(validationPage, libraryName, skylineDoc, librarySize,
+                Collections.emptyList(), rawSources,
+                Collections.emptyList(), peptideIdSources);
+
+        // Upload the .speclib file.
+        log("Uploading file " + peptideIdSources.get(0) + " to folder - " + folderName);
+        uploadToRawFiles(testFilesFolder + "/" + peptideIdSources.get(0));
+        log("Running data validation job. " + peptideIdSources.get(0) + " shoul");
```
labkey-jeckels:
```suggestion
```

panoramapublic/webapp/PanoramaPublic/js/pxValidation.js
```diff
@@ -79,7 +79,11 @@ renderFileStatus: function renderFileStatus(values, isSampleFile, link) {
     if (values.found === true) status = "FOUND";
     else if (values.ambiguous === true) status = "AMBIGUOUS"
```
labkey-jeckels:
```suggestion
    else if (values.ambiguous === true) status = "AMBIGUOUS";
```

panoramapublic/src/org/labkey/panoramapublic/speclib/LibSourceFile.java
```diff
@@ -81,4 +81,14 @@ public boolean isMaxQuantSearch()
...
     }
+
+    public static String DIANN_REPORT_TSV_PLACEHOLDER = "DIA-NN report file";
+
+    // These are some of the column headers that we expect to see in a DIA-NN report TSV file
+    public static List<String> DIANN_REPORT_EXPECTED_HEADERS = List.of("File.Name", "Run", "Protein.Group", "Protein.Ids", "Protein.Names");
+
+    public boolean isDiannSearch()
+    {
+        return (hasIdFile() && getIdFile().endsWith(".speclib"));
```
labkey-jeckels: In some places file extension checks are case-insensitive. Does that matter here?

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/speclib/LibSourceFile.java
vagisha: Perhaps.  I've changed it to case-insensitive check.

PR #522, opened 2025-03-14
Use PageFlowUtil.jsString to escape container path

vagisha:
#### Rationale
Properly escape folder names passed as parameter to JS function, otherwise names containing single quotes cause JavaScript error.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
@@ -162,7 +162,10 @@ public class ExperimentAnnotationsTableInfo extends FilteredTable<PanoramaPublic
                                             .at(src, PageFlowUtil.staticResourceUrl("_images/plus.gif"))),
                                     HtmlString.NBSP)
                                     .appendTo(out);
-                            pageConfig.addHandler(spanId, "click", "viewExperimentDetails(this,'" + container.getPath() + "', '" + id + "','" + detailsPage + "')");
+                            pageConfig.addHandler(spanId, "click", "viewExperimentDetails(this,"
+                                    + PageFlowUtil.jsString(container.getPath())
+                                    + ", '" + id + "', "
+                                    + PageFlowUtil.jsString(detailsPage) + ")");
                         }
                         super.renderGridCellContents(ctx, out);
                     }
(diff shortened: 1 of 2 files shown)


panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
```diff
@@ -162,7 +162,10 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                                             .at(src, PageFlowUtil.staticResourceUrl("_images/plus.gif"))),
                                     HtmlString.NBSP)
                                     .appendTo(out);
-                            pageConfig.addHandler(spanId, "click", "viewExperimentDetails(this,'" + container.getPath() + "', '" + id + "','" + detailsPage + "')");
+                            pageConfig.addHandler(spanId, "click", "viewExperimentDetails(this,"
+                                    + PageFlowUtil.jsString(container.getPath())
+                                    + ", '" + id + "', "
```
labkey-jeckels: Might as well handle them all the same way.

```suggestion
                                    + ", " + PageFlowUtil.jsString(id) + ", "
```

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
vagisha: 'id' is a non-null Integer so does not need to be quoted, I think. 

I added some HTML-escaping in related code.  If you could take another look that would be great.

labkey-jeckels (approved):

panoramapublic/webapp/PanoramaPublic/js/dropDownUtil.js
```diff
@@ -66,12 +66,13 @@ viewExperimentDetails = function (obj, experimentContainer, id, detailsPageURL)
         var results;
         if(object.rows[rowNum][type] != null)
         {
-            if(object.rows[rowNum][type].length > 500)
+            let description = object.rows[rowNum][type];
+            if(description.length > 500)
             {
-                results = object.rows[rowNum][type].substring(0,500)+"<a href='"+detailsPageURL+"'>...more.</a>";
+                results =  LABKEY.Utils.encodeHtml(description.substring(0,500)) +"<a href='"+ LABKEY.Utils.encodeHtml(detailsPageURL) +"'>...more.</a>";
```
labkey-jeckels:
```suggestion
                results =  LABKEY.Utils.encodeHtml(description.substring(0,500)) +"<a href=\""+ LABKEY.Utils.encodeHtml(detailsPageURL) +"\">...more.</a>";
```

panoramapublic/src/org/labkey/panoramapublic/query/ExperimentAnnotationsTableInfo.java
labkey-jeckels: You're right. I somehow thought it was a string.

PR #524, opened 2025-03-19
Switch to tracking NextFlow invocation counts per-job

labkey-jeckels:
#### Rationale
Admins and users will find it easier if we track NextFlow invocations on a per-job basis, instead of a globally across all NextFlow jobs.

#### Changes
* Create a nextflow schema and table table to track invocation counts
* Change logging to more clearly differentiate between the initial queuing and subsequent execution-based messages

### DIFF
--- a/nextflow/src/org/labkey/nextflow/pipeline/NextFlowRunTask.java
+++ b/nextflow/src/org/labkey/nextflow/pipeline/NextFlowRunTask.java
@@ -2,9 +2,6 @@ package org.labkey.nextflow.pipeline;
 
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
-import org.labkey.api.data.ContainerManager;
-import org.labkey.api.data.DbSequence;
-import org.labkey.api.data.DbSequenceManager;
 import org.labkey.api.exp.XarFormatException;
 import org.labkey.api.pipeline.AbstractTaskFactory;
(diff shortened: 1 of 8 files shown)


nextflow/src/org/labkey/nextflow/pipeline/NextFlowRunTask.java
```diff
@@ -54,9 +49,9 @@ public NextFlowRunTask(Factory factory, PipelineJob job)
 
         // NextFlow requires a unique job name for every execution. Increment a counter to append as a suffix to
         // ensure uniqueness
-        long invocationCount = INVOCATION_SEQUENCE.next();
-        INVOCATION_SEQUENCE.sync();
-        NextFlowPipelineJob.LOG.info("Starting to execute NextFlow: {}", getJob().getJsonJobInfo(invocationCount));
+        NextFlowManager.get().incrementInvocationCount(getJob());
```
ankurjuneja, line 52: should this be in try catch?

ankurjuneja (approved):

nextflow/src/org/labkey/nextflow/pipeline/NextFlowRunTask.java
labkey-jeckels: I don't think that would add anything. We need to increment the number whether the execution succeeds or not. I need the new value for the logging call on the next line, so I can't really delay it either.

PR #544, opened 2025-04-21
Create a post on Bluesky when a dataset is made public

vagisha:
#### Rationale
We would like to post a message to the Bluesky social media platform after a dataset is made public on Panorama Public.

#### Changes
- Added configuration page for Bluesky Settings under Panorama Public Admin Console.  This can be used to configure the following:
  - Credentials for both primary and test Bluesky accounts
  - Provide the Bluesky API endpoint URLs for login, posting, and image upload
  - Specify default hashtags for posts
  - Upload a default Panorama logo (overridden by a dataset’s catalog entry image, if available)
  - "Auto‑post" toggle (off by default) to enable automatic posting when a submitter makes a dataset public
- Added "Post to Bluesky" menu item to the TargetedMSExperimentWebPart.  This allows admins to create a post for the dataset to either the primary or test account.
- When “auto‑post” is enabled, a Bluesky post is created immediately after a dataset is made public
- Store generated Bluesky post URLs to a property store, linked to each dataset’s short URL


panoramapublic/src/org/labkey/panoramapublic/bluesky/PanoramaPublicLogoResourceType.java
```diff
@@ -0,0 +1,35 @@
...
+        return getClass().getName();
+    }
+
+    @Override
+    public void addWhereSql(SQLFragment sql, String parentColumn, String documentNameColumn)
+    {
+        sql.append(parentColumn).append(" IN (SELECT EntityId FROM ")
+                .append(CoreSchema.getInstance().getTableInfoContainers(), "c").append(")")
+                .append(" AND (")
+                .append(documentNameColumn).append(" LIKE '").append(PanoramaPublicLogoManager.LOGO_FILE_PREFIX).append("%' )");
```
labkey-jeckels: This constant is unlikely to change to a dangerous value, but safer to use `SQLFragment.appendStringLiteral()`

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -5214,6 +5434,241 @@ public void setDoi(String doi)
...
+            BlueskySettings settings = BlueskySettingsManager.getSettings();
+            try
+            {
+                // Post to Bluesky
+                _blueskyAtUri = BlueskyApiClient.getInstance().createPost(_expAnnot, settings, form.isTestAccount());
+            }
+            catch (BlueskyException e)
+            {
+                _exception = e;
+                LOG.error(e);
```
labkey-jeckels: I think this will just get you the `toString()`. Maybe that's enough here, but you might want the two-argument version which will get you the full stack trace

labkey-jeckels (approved):

PR #554, opened 2025-06-08
Minor auto-refactor code cleanup on a massive scale

labkey-jeckels:
#### Rationale
We can cleanup tens of thousands of warnings across our codebase with IntelliJ's autorefactors. In some cases, this adopts newer, leaner syntax. In others, it simply removes code that's not serving any purpose.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6742

#### Changes
- Auto-refactors
  - Eliminate redundant toString()
  - Eliminate redundant cast
  - Empty JavaDoc annotations
  - Member variable can be final
  - length() and size() -> isEmpty()
  - Add missing @Override
  - Class can be static
  - Remove unused imports
  - toArray() passed array length
  - Remove unnecessary semicolon
  - Explicit type syntax can be replaced by <>
  - Redundant access modifiers
  - Remove redundant initializer
  - "".equals() -> isEmpty()
  - StringBuilder can be replaced by String
  - Fix misordered arguments to assertEquals()
  - StringUtils.defaultString() - Objects.toString()
  - Cast can be replaced with pattern variable
  - Collapse identical catch blocks
  - Type may be primitive
- Manual fixups
  - Delete unused GWT code
  - Improve generics
  - new UnexpectedException() -> UnexpectedException.wrap()

### DIFF
--- a/SkylineToolsStore/src/org/labkey/skylinetoolsstore/RatingManager.java
+++ b/SkylineToolsStore/src/org/labkey/skylinetoolsstore/RatingManager.java
@@ -87,7 +87,7 @@ public class RatingManager
             for (SkylineTool tool : tools)
                 ratings.addAll(Arrays.asList(getRatingsByToolId(tool.getRowId())));
         }
-        return ratings.toArray(new Rating[ratings.size()]);
+        return ratings.toArray(new Rating[0]);
     }
 
     public boolean userLeftRating(String toolLsid, User user)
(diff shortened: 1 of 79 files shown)


vagisha (approved):
Thanks for the cleanup! I’ve skimmed through the changes - they all look minor. Let me know if you think any of this needs manual testing.

labkey-jeckels:
> Thanks for the cleanup! I’ve skimmed through the changes - they all look minor. Let me know if you think any of this needs manual testing.

I don't think so. They should all be safe changes that IntelliJ was able to auto-refactor, or very minor cleanup I applied directly.

PR #579, opened 2025-11-14
Remove public account details from the "Download Data" webpart

vagisha:
#### Changes
- Removed public account details from the "Download Data" webpart
- Added a note about contacting the PanoramaWeb team before initiating very large downloads via WebDAV


labkey-jeckels (approved):
Do you want to also prompt people to create an account if they don't have one?

vagisha:
> Do you want to also prompt people to create an account if they don't have one?

I updated the text.  Thanks.

PR #584, opened 2025-11-24
Continue adopting FileLike instead of File or Path

labkey-jeckels:
#### Rationale
We can continue to strengthen our protections against path traversal attacks and start phasing out deprecated and overloaded methods that take or return Files or Paths. 

### Related Pull Requests
- https://github.com/LabKey/platform/pull/7211

#### Changes
- Eliminate some overloaded, deprecated versions of methods taking `File` or `Path`
- Convert `LocalDirectory` and `WorkDirectory` to `FileLike`
- Use `FileLike` more consistently for transform scripts

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
@@ -4723,7 +4723,7 @@ public class PanoramaPublicController extends SpringActionController
             }
             else
             {
-                return FileSystemLike.wrapFile(fileRoot.resolve(fileName));
+                return FileSystemLike.wrapFile(fileRoot).resolveChild(fileName);
             }
         }
     }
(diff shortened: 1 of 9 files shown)


panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -4722,7 +4723,7 @@ private static File getLocalFile(Container container, String fileName) throws Px
             }
             else
             {
-                return fileRoot.resolve(fileName).toFile();
+                return FileSystemLike.wrapFile(fileRoot.resolve(fileName));
```
labkey-matthewb: is Path.resolve() path traversal safe?
Could do FileSystemLike.wrapFile(fileRoot).resolveChild(fileName));

labkey-matthewb (approved):

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
labkey-jeckels: Updated.

PR #596, opened 2026-01-22
Fix failure in PanoramaWebPublicSearchTest due to invalid URL parameters

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/bt310/3814947?buildTab=tests&status=failed&suite=org.labkey.test.Runner%3A+&package=org.labkey.test.tests.panoramapublic&class=PanoramaWebPublicSearchTest

We're generating invalid URLs with empty GET parameters. Example snippet:

`&Targeted%20MS%20Experiment%20List.authors~containsoneof=Doe&=&dataRegionName=Targeted%20MS%20Experiment%20List`

Tomcat 11 has gotten stricter about validating the URL structure.

#### Changes
* Don't blindly append `&` when we may not need the separator

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
+++ b/panoramapublic/src/org/labkey/panoramapublic/view/search/panoramaWebSearch.jsp
@@ -499,7 +499,6 @@
                                ];
                             }
 
-                            debugger;
                             let wp = new LABKEY.QueryWebPart({
                                 renderTo: 'webpart_'+ expWebpart[0].webPartId,
                                 title: 'Panorama Public Experiments',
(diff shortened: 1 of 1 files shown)


labkey-adam:
@labkey-jeckels shouldn't this use client-side helpers to build URLs to avoid these kinds of problems?

labkey-jeckels:
> @labkey-jeckels shouldn't this use client-side helpers to build URLs to avoid these kinds of problems?

Yes, it should. However, the code I modified is using GET parameter-style values and appending them to the hash part of the URL by appending to window.location.href.  ActionURL isn't built to do this. As much as I'd like to refactor a good chunk of this file, I didn't want to tackle it.

ankurjuneja (approved):

PR #599, opened 2026-02-05
Added Retrain All feature and improved menu styling

brendanx67:
## Summary

* Added RetrainAllAction with Reset/Incremental modes for bulk training data management
* Reset mode rebuilds training data from scratch using recent clean runs
* Incremental mode uses rolling window to replace oldest runs with newer ones
* Added Min/Max runs parameters (defaults 5/20) with 1.5x lookback period
* Redesigned menu with active tab highlighting (rounded corners) and gold hover
* All pages now set activeTab attribute for consistent menu state

### Retrain All Feature
- **Reset mode**: Deletes all existing training data and rebuilds from the most recent clean runs within a lookback period (1.5x max runs = 30 days by default)
- **Incremental mode**: Rolling window that replaces oldest training runs with newer clean runs while maintaining continuity
- **Min runs** (default 5): Minimum clean runs required to create training data for a computer
- **Max runs** (default 20): Maximum training runs per computer; also determines lookback period
- Clean run criteria: 0 failures, 0 leaks, no hang, full duration, not flagged, passedtests > 0

### Menu Improvements
- Active tab highlighting with rounded corners and semi-transparent background
- Gold hover color matching original UW branding
- Consistent activeTab attribute across all pages
- Improved CSS organization using classes instead of inline styles

## Test plan

- [x] Tested Reset mode on Performance Tests container
- [x] Verified lookback period filters out ancient data
- [x] Verified computers with partial data (>= minRuns but < maxRuns) appear with training data but inactive
- [x] Verified menu highlighting works on all pages

### DIFF
--- a/testresults/src/org/labkey/testresults/TestResultsController.java
+++ b/testresults/src/org/labkey/testresults/TestResultsController.java
@@ -111,6 +111,8 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


testresults/src/org/labkey/testresults/TestResultsController.java
```diff
@@ -1037,6 +1039,261 @@ public Object execute(Object o, BindException errors)
         }
     }
 
+    @RequiresSiteAdmin
+    public static class RetrainAllAction extends MutatingApiAction<Object>
```
vagisha: This action extends `MutatingApiAction<Object>` but manually parses request parameters. Instead, you could define a form class and use `MutatingApiAction<RetrainAllForm>` to get automatic parameter binding, type safety, and default values.
Manual parsing like this:
```
String minRunsParam = req.getParameter("minRuns");
if (minRunsParam != null && !minRunsParam.isEmpty())
{
    try { minRuns = Integer.parseInt(minRunsParam); }
    catch (NumberFormatException ignored) { }
}
```
could be replaced with: 

```Integer minRuns = form.getMinRuns();```

File: testresults/src/org/labkey/testresults/TestResultsController.java [hunk 2]
```diff
@@ -1037,6 +1039,261 @@ public Object execute(Object o, BindException errors)
...
+    {
+        @Override
+        public Object execute(Object o, BindException errors)
```
vagisha: This method is too long.  Consider refactoring.  All the SQL code could be moved to the `TestResultsManager` class.  That's the convention I normally follow.

testresults/src/org/labkey/testresults/view/rundown.jsp
```diff
@@ -92,6 +92,7 @@
 </style>
 
 <div id="container">
+<% request.setAttribute("activeTab", "overview"); %>
```
vagisha: Tab identifiers like "run", "overview", "trainingdata" are hardcoded across multiple JSPs. They could be added as constants in `TestResultsController`
```
public static class TabNames
{
        public static final String OVERVIEW = "overview";
        public static final String USER = "user";
        public static final String RUN = "run";
        public static final String LONGTERM = "longterm";
        public static final String FLAGS = "flags";
        public static final String TRAINING_DATA = "trainingdata";
        public static final String ERRORS = "errors";
}
 ```
In the JSP

```
<%@ page import="static org.labkey.testresults.TestResultsController.TabNames" %>

<% request.setAttribute("activeTab", TabNames.OVERVIEW); %>
```

vagisha (approved):

vagisha:
I left a few comments.  We should refactor the `testresults`, `signup `and `SkylineToolsStore `modules to use standard LabKey base action classes (`FormViewAction`,  `FormHandlerAction`, etc.), and other standard LabKey patterns as well as some basic tests.  This could be a good exercise for Claude Code?

brendanx67:
> I left a few comments. We should refactor the `testresults`, `signup `and `SkylineToolsStore `modules to use standard LabKey base action classes (`FormViewAction`, `FormHandlerAction`, etc.), and other standard LabKey patterns as well as some basic tests. This could be a good exercise for Claude Code?

Let me know what you think of the edits I had Claude Code do. Yes, I think you could be pretty successful with having Claude Code make the other MacCoss lab modules more standard. Right now, I just want the ability to press a single button and update the training data with a new snapshot based on the heuristics I was performing manually.

testresults/src/org/labkey/testresults/view/rundown.jsp
brendanx67: Okay. Did some of this clean-up.

testresults/src/org/labkey/testresults/TestResultsController.java
brendanx67: Okay. Had Claude Code refactor to move queries to TestResultsManager.

testresults/src/org/labkey/testresults/TestResultsController.java
brendanx67: Had Claude Code do this as suggested. There is now a form class.

PR #566, opened 2025-08-18
Send reminders to submitters of private data on Panorama Public

vagisha:
#### Rationale
Submitters often forget to make their data public even after the associated paper is published.  We should send them reminders periodically.

#### Changes
- Send automated reminders to submitters of private datasets. 
- Reminders can be enabled, and properties such as "delay until first reminder", "reminder frequency" and "extension duration" can be configured via the Panorama Public Admin Console,  
- The Admin Console also provides an option to send reminders manually.
- Submitters can request an extension of the private data status or request deletion of their data from Panorama Public. 
- Links to request extension or deletion are included in the reminder message. 
- Extension and deletion requests by the submitter are posted to the data message thread for record keeping. 
- Escape tilde (~) when generating notification messages. In the LabKey Markdown flavor, text between single tildes (e.g., \~strikethrough\~) as well as double tildes is rendered as ~strikethrough~.

PR #606, opened 2026-03-10
Search NCBI for papers associated with private datasets

vagisha:
#### Rationale
Enhance the Private Data Reminder system to automatically detect publications associated with private datasets by searching PubMed Central and PubMed. When a publication is found, the reminder message includes the citation and encourages the submitter to make their data public.

#### Related Pull Requests
- https://github.com/LabKey/MacCossLabModules/pull/566

#### Changes
- Added a setting in the admin console (_Private Data Reminder Settings_) to enable publication search
- Added `NcbiPublicationSearchService` that searches for publications in
  - PubMed Central (full-text available): search terms ProteomeXchange Id, Panorama Public short URL, Panorama Public DOI.  Author and title verified.
  - PubMed - fallback if no matches found in PMC; search author and title only since PubMed does not have full-text articles
  - Preprints (e.g. biorxiv, medrxiv etc.) are filtered out
  - Results are prioritized by number of matched search terms (e.g. PX ID, Panorama link) as well as paper publication date proximity to data submission date
- Publication search happens automatically in the `PrivateDataReminderJob`  when enabled in the admin console
- Publications for a single dataset can also be searched through the **Search Publications** menu item in the _TargetedMS Experiment_ webpart menu
- If a publication match is found during the `PrivateDataReminderJob`,  instead of the usual reminder message, a "publication found" message is posted to the submitter
- Submitter can dismiss suggested publication by clicking the **Dismiss Publication Suggestion** link in the message. 
- `DatasetStatus` caches publication result to avoid repeated calls to NCBI's EUtils endpoints
- Moved NCBI citation lookup methods to `NcbiPublicationSearchServiceImpl`
- **Tests**
   - Unit tests added in `NcbiPublicationSearchServiceImpl`. 
   - Added Selenium test `PublicationSearchTest`. 
     - Test uses a mock NCBI service on TeamCity (mocking only the outbound HTTP requests) so all real search and filtering logic is exercised without live API calls.

[private-data-reminders-overview.md](https://github.com/user-attachments/files/25857390/private-data-reminders-overview.md)
[SPEC-SUMMARY.md](https://github.com/user-attachments/files/25857424/SPEC-SUMMARY.md)


panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+        return executeSearch(query, "pubmed", log);
+    }
+
+    /**
+     * Execute search using NCBI ESearch API
+     */
+    private List<String> executeSearch(String query, String database, Logger log)
+    {
+        String encodedQuery = URLEncoder.encode(query, StandardCharsets.UTF_8);
+        String url = ESEARCH_URL +
```
labkey-jeckels, line 306: These parameters should be URL-encoded. `encodedQuery` is already handled but unless `database` and others are known to the safe, they should be encoded.

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+
+    /**
+     * Fetch metadata using NCBI ESummary API (batch request)
+     */
+    private Map<String, JSONObject> fetchMetadata(Collection<String> ids, String database, Logger log)
+    {
+        if (ids.isEmpty()) return Collections.emptyMap();
+
+        String idString = String.join(",", ids);
+        String url = ESUMMARY_URL +
```
labkey-jeckels, line 401: URL encoding

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+        String title = expAnnotations.getTitle();
+
+        if (StringUtils.isBlank(firstName) || StringUtils.isBlank(lastName) || StringUtils.isBlank(title))
+        {
+            log.info("Cannot perform PubMed fallback - missing author or title information");
+            return Collections.emptyList();
+        }
+
+        // Search PubMed: "LastName FirstName[Author] AND Title NOT preprint[Publication Type]"
+        String query = String.format("%s %s[Author] AND %s NOT preprint[Publication Type]",
```
labkey-jeckels, line 585: I don't know the PubMed search syntax very well, but any need to do escaping here?

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+     * Normalize title for comparison: strip HTML tags, strip diacritics, lowercase,
+     * replace punctuation with space (so "data-independent" becomes "data independent"),
+     * and normalize whitespace.
+     */
+    static String normalizeTitle(String title)
+    {
+        if (title == null) return "";
+
+        return stripDiacritics(title.toLowerCase())
+            .replaceAll("<[^>]+>", " ")    // Strip HTML/XML tags
```
labkey-jeckels: If we need to strip tags, do we also need to HTML-decode?

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+     */
+    private static void rateLimit()
+    {
+        try
+        {
+            Thread.sleep(RATE_LIMIT_DELAY_MS);
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
```
labkey-jeckels, line 915: I think this state will already be set by virtue of the exception, but shouldn't be harmful either.

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
```diff
@@ -0,0 +1,1391 @@
...
+        }
+    }
+
+    /**
+     * Wrap string in quotes for exact match search
+     */
+    private static String quote(String str)
+    {
+        return "\"" + str + "\"";
```
labkey-jeckels: Should this escape quotes within the string?

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -10566,6 +10771,298 @@ public void setShortUrlEntityId(String shortUrlEntityId)
...
+            if (StringUtils.isBlank(form.getPublicationId()))
+            {
+                errors.reject(ERROR_MSG, "No publication was selected");
+            }
+        }
+
+        @Override
+        public boolean handlePost(NotifySubmitterForm form, BindException errors) throws Exception
+        {
+            ExperimentAnnotations exptAnnotations = ExperimentAnnotationsManager.get(form.getId());
```
labkey-jeckels, line 10934: Should this also check that the container matches?

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -6946,7 +6951,6 @@ public ModelAndView getView(PublicationDetailsForm form, boolean reshow, BindExc
                 return new SimpleErrorView(errors);
             }
 
-            form.setPubmedId(_copiedExperiment.getPubmedId());
```
labkey-jeckels, line 6949: Was this intentional? If so, great. If not, we're not propagating the ID anymore.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
```diff
@@ -10566,6 +10771,298 @@ public void setShortUrlEntityId(String shortUrlEntityId)
...
+            List<User> notifyUsers = new ArrayList<>();
+            notifyUsers.add(submitter);
+            if (exptAnnotations.getLabHeadUser() != null)
+            {
+                notifyUsers.add(exptAnnotations.getLabHeadUser());
+            }
+
+            PanoramaPublicNotification.postPrivateDataReminderMessage(
+                    journal, submission, exptAnnotations, submitter, getUser(), notifyUsers,
+                    _announcement, _announcementsContainer, getUser(), selectedMatch);
```
labkey-jeckels: Should this and subsequent updates in this method be transacted?

labkey-jeckels (approved):

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
vagisha: They are string constants and are safe, but URL-encoding them is correct practice. Extracted a `buildCommonParams()` helper that encodes all three common parameters (db, tool, email)

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
vagisha: Done

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
vagisha: There is no escape mechanism for special characters as far as I can tell. Per  https://pubmed.ncbi.nlm.nih.gov/help/ the recommendation is to remove special characters.   Added `stripQuerySpecialChars()` to strip `[]()\"`.

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
vagisha: Good point.  So far I have only seen formatting tags (`<i>`, `<sub>` etc.).  But I have added `StringEscapeUtils.unescapeHtml4()` just in case.

panoramapublic/src/org/labkey/panoramapublic/ncbi/NcbiPublicationSearchServiceImpl.java
vagisha: This is only called for PX ID, Panorama short URL and DOIs which will not contain quotes.  But I've updated the method to strip quotes defensively.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: Ah, yes.  Added a call to `ensureCorrectContainer()`.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: No!  However, the form can arrive pre-populated via the URL when the user clicks the "Make Public" link in a notification. Fixed the code to fall back to the copied experiment's PubMed ID when the form doesn't already have one.

panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicController.java
vagisha: Yes, they should be.  Fixed.

PR #646, opened 2026-06-21
TestResults module bugfixes

vagisha:
#### Rationale
Fixes several bugs in the testresults module found while writing Selenium tests, plus a folder-scoping gap in how
runs are looked up by id:

* Removing a computer's last training run left a stale statistics row on the Training Data page.
* Deleting a run that had handle leaks or was in the training set failed on a foreign-key constraint.
* The failure chart threw a JavaScript error when the selected range had no dates.
* Several actions lookup a run by its runId without filtering on its container column, so a run lookup by id could read or modify a run in another folder by guessing its id.
* Out-of-range dates like 13/45/2026 were silently accepted (the shared date parser was lenient), and the parser was a non-thread-safe static instance.

#### Related Pull Requests
* #645 

#### Changes

* Training stats: extracted `recomputeUserData`(scope, userid, container), which recomputes a user's stats from their remaining training runs and deletes the `UserData `row when none remain. Called by `TrainRunAction `and `DeleteRunAction`.
* `DeleteRunAction`: delete all child rows (now including handleleaks and trainruns) before the parent run, then refresh the owner's training stats.
* `TrainRunAction`: verify the run exists before recompute on the force path too.
* Folder scoping: added `getRunInContainer(runId, container)`; `trainRun`, `deleteRun`, `flagRun`, `viewLog`, `viewXml `and the flagged-runs list now reject or omit runs from other folders. `showRun `and `TrainingDataViewAction `were already scoped.
* `failureDetail.jsp`: set the chart's date-axis bounds only when dates exist.
* Date parsing: parse MM/dd/yyyy strictly and reject invalid dates; use a fresh formatter per call instead of one shared static instance.

## Testing

Added new tests for strict dates, deleting a run with child rows, the recompute update branch (mean and stddev), the trainRun force-path error, and cross-folder run access (testRunAccessIsContainerScoped).

### DIFF
--- a/testresults/src/org/labkey/testresults/view/failureDetail.jsp
+++ b/testresults/src/org/labkey/testresults/view/failureDetail.jsp
@@ -266,6 +266,19 @@ $(document).ready(function() {
     const problemData = <%=json(problemData, 0)%>;
 
     // Generate date chart.
+    // Only set explicit timeseries min/max when there are dates to plot. With an empty
+    // dates array, dates[0] and dates[length - 1] are undefined and c3/d3 throws while
+    // building the x-axis scale (e.g. when the active date range excludes all sample runs).
+    const chartDates = problemData.graphData.dates;
+    const xAxis = {
+        type: 'timeseries',
+        localtime: false,
+        tick: { fit: true, format: '%m/%d' }
+    };
+    if (chartDates.length > 0) {
+        xAxis.min = chartDates[0];
+        xAxis.max = chartDates[chartDates.length - 1];
+    }
     let dateChart = c3.generate({
         bindto: '#failGraph',
         data: {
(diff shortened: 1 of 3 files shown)


testresults/src/org/labkey/testresults/view/failureDetail.jsp
```diff
@@ -266,6 +266,19 @@ $(document).ready(function() {
     const problemData = <%=json(problemData, 0)%>;
 
     // Generate date chart.
+    // Only set explicit timeseries min/max when there are dates to plot. With an empty
+    // dates array, dates[0] and dates[length - 1] are undefined and c3/d3 throws while
+    // building the x-axis scale (e.g. when the active date range excludes all sample runs).
+    const chartDates = problemData.graphData.dates;
```
ankurjuneja:
```suggestion
    const chartDates = problemData?.graphData?.dates;
```

ankurjuneja (approved):

PR #652, opened 2026-06-29
Address reported security issues in the LINCS module

vagisha:
#### Rationale
A security review of the LINCS module identified six issues, addressed here:
* High - path traversal (arbitrary file read + delete) in DownloadCustomGCTReportAction
* Medium / Low - PSP job lookups not container-scoped (cross-folder data access)
* Low (CWE-319) - Clue/PSP API key transmittable over http
* Low - credential / config changes not audited

#### Changes
**Path traversal (High) - DownloadCustomGCTReportAction**
* Validate the requested `fileName` with `FileUtil.isAllowedFileName` and confirm the resolved path stays inside the container's GCT directory.

**Cross-container access (Medium / Low) - PSP job lookups**
* `LincsManager.getLincsPspJob` / `getLincsPspJobForRun` now require a `Container` and filter on it. 
   The PSP-job detail/status/update actions can no longer reach jobs in other  folders. 
   All callers updated (3 controller actions, SubmitPspJobAction, and 2 LincsDataTable display columns).

**Cleartext credential transmission (Low, CWE-319) - Clue/PSP server URI**
* Require `https` for the Clue/PSP server URI: rejected at save time in `ManageLincsClueCredentials.validateCommand`, 
   and as defense in depth in `LincsPspUtil.getPspEndpoint` before the API key is sent as a request header.
  The point-of-use check also blocks endpoints saved as `http://` before this fix.
* Added `<labkey:errors/>` to `manageClueCredentials.jsp` so the save-time rejection is actually shown.

**Missing audit trail (Low) - credential / config changes**
* `ManageLincsClueCredentials.handlePost` and `CromwellConfigAction.handlePost`  now write an audit event (container and acting user, no secret values) when the PSP/Clue credentials or Cromwell config change.

### DIFF
--- a/lincs/src/org/labkey/lincs/DocImportListener.java
+++ b/lincs/src/org/labkey/lincs/DocImportListener.java
@@ -41,11 +41,11 @@ public class DocImportListener implements ExperimentListener, SkylineDocumentImp
             return;
         }
 
-        ITargetedMSRun tRun = TargetedMSService.get().getRunByFileName(run.getName(), run.getContainer());
+        ITargetedMSRun tRun = TargetedMSService.get().getRunByFileName(run.getName(), c);
         if(tRun != null)
         {
             // Delete saved entries in lincs.lincspspjob table for this runId
-            LincsManager.get().deleteLincsPspJobsForRun(tRun.getId());
+            LincsManager.get().deleteLincsPspJobsForRun(tRun.getId(), c);
         }
 
         // Get the file root for the container
--- a/lincs/src/org/labkey/lincs/LincsController.java
+++ b/lincs/src/org/labkey/lincs/LincsController.java
@@ -733,8 +733,7 @@ public class LincsController extends SpringActionController
 
             Path gctDir = getGCTDir(getContainer()).normalize();
             Path downloadFile = gctDir.resolve(form.getFileName()).normalize();
-            // Defense in depth: ensure the resolved path is still inside the GCT directory before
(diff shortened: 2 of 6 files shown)


ankurjuneja (approved):
Thanks @vagisha , here is one finding from claude code review using LK skills that I think is relevant here:

Finding #1 - [Severity: Medium] — Consistency / IDOR gap — LincsManager.deleteLincsPspJobsForRun(long runId)

▎ Issue: The PR added a Container parameter to getLincsPspJobForRun and getLincsPspJob, but deleteLincsPspJobsForRun(long runId) (visible just below in the same file) was left filtering by RunId only, with no container filter.
▎ Why it matters: If this method is ever reachable with an attacker-influenced runId, it would delete PSP jobs across container boundaries — the same class of issue the rest of the PR closes. It's likely only invoked internally on run-deletion events (where the run is already container-scoped), which would make it safe, but that should be confirmed.
▎ Suggestion: Verify the call sites. If any path takes a request-supplied runId, add SimpleFilter.createContainerFilter(container) here too for symmetry.

vagisha:
Thanks, @ankurjuneja. `deleteLincsPspJobsForRun` has exactly one caller, and the `runId `it passes is container-scoped. The `runId` is never supplied by a user request. So the current code is safe in practice, but I added container-filtering to be consistent - 289226aeaa3fbec6a514fad292b01a0857362d60

PR #653, opened 2026-06-29
nextflow security fixes

ankurjuneja:
#### Rationale
Two security/robustness issues were found in NextFlowController:                                                                                                                                                                  
                                                                                                                                                                                                                                    
  1. Function-level authorization gap (BeginAction.handlePost). BeginAction is gated only by ReadPermission, and the site-admin check (hasSiteAdminPermission()) lived solely in getView() — the UI only renders the enable/disable 
  toggle for site admins. The POST handler that actually writes the per-folder enabled state had no privilege re-check, so any folder reader could POST directly to the action and toggle NextFlow integration for the folder.      
  Because isEnabled() walks up the container tree, child containers inherit that state, widening the impact. The editing capability is intended for site admins only, so the invariant needed to be re-established on the write     
  path.                                                                                                                                                                                                                             
  2. NullPointerException on the run path (NextFlowRunAction). NextFlowManager.getConfiguration() returns null when no NextFlow configuration has been saved. Both getView() and handlePost() immediately dereferenced the result   
  (config.getNextFlowConfigFilePath()), producing a 500 error for an editor working in a container where NextFlow is enabled but not yet configured. Users should get a clear message instead of a server error.                    

#### Related Pull Requests
* <!-- list of links to related pull requests (replace this comment) -->

#### Changes
* NextFlowController.BeginAction.handlePost                                                                                                                                                                                         
  - Added a hasSiteAdminPermission() check that throws UnauthorizedException before writing the enabled state, matching the gate already present in getView(). Enabling alone still does not grant run rights — NextFlowRunAction   
  continues to require InsertPermission. 
  - NextFlowRunAction.handlePost: wrapped the FileUtil.appendPath call (which already enforces directory containment / rejects .. traversal) in a try/catch so an invalid configFile yields a user-facing "Invalid config file"     
  error instead of an unhandled 500.                                                                                                                                                                                           
                                                                                                                                                                                                                                    
* NextFlowController.NextFlowRunAction                                                                                                                                                                                              
  - validateCommand: reject with "NextFlow has not been configured" when getConfiguration() is null (guards the POST path).                                                                                                         
  - getView: added a config != null guard, falling through to the existing "Couldn't find NextFlow config file(s)" message.                                                                                                         
  - handlePost: added a defensive null/empty-path check (it re-fetches the configuration independently), rejecting with a user-facing error instead of throwing.

### DIFF
--- a/nextflow/src/org/labkey/nextflow/NextFlowController.java
+++ b/nextflow/src/org/labkey/nextflow/NextFlowController.java
@@ -45,6 +45,7 @@ import org.springframework.validation.Errors;
 import org.springframework.web.servlet.ModelAndView;
 
 import java.io.File;
+import java.nio.file.InvalidPathException;
 import java.util.Arrays;
 import java.util.List;
 
@@ -325,7 +326,17 @@ public class NextFlowController extends SpringActionController
                 return false;
             }
             File configDir = new File(config.getNextFlowConfigFilePath());
-            File configFile = FileUtil.appendPath(configDir, Path.parse(form.getConfigFile()));
+            File configFile;
+            try
+            {
+                // appendPath normalizes and enforces that the resolved path stays within configDir, rejecting traversal
+                configFile = FileUtil.appendPath(configDir, Path.parse(form.getConfigFile()));
+            }
+            catch (InvalidPathException e)
+            {
+                errors.reject(ERROR_MSG, "Invalid config file");
+                return false;
+            }
             if (!configFile.exists())
             {
                 errors.reject(ERROR_MSG, "Config file does not exist");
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

nextflow/src/org/labkey/nextflow/NextFlowController.java
```diff
@@ -311,8 +320,23 @@ public boolean handlePost(AnalyzeForm form, BindException errors) throws Excepti
...
+                errors.reject(ERROR_MSG, "NextFlow has not been configured");
+                return false;
+            }
             File configDir = new File(config.getNextFlowConfigFilePath());
-            File configFile = FileUtil.appendPath(configDir, Path.parse(form.getConfigFile()));
+            File configFile;
+            try
+            {
+                // appendPath normalizes and enforces that the resolved path stays within configDir, rejecting traversal
+                configFile = FileUtil.appendPath(configDir, Path.parse(form.getConfigFile()));
```
vagisha, line 338: `Path.parse` could throw a NPE if `getConfigFile()` returns `null`.  Should we add a `StringUtils.isBlank(form.getConfigFile())` guard before this `try/catch`?

vagisha (approved):

PR #662, opened 2026-07-11
Restrict the modification query grids to the current folder tree

vagisha:
#### Rationale

The modification web part title links to the full query grid, where an "All Folders" filter could be applied, leading to very slow queries. Restricting the Folder Filter menu is not enough because a hand-built `query.containerFilterName=AllFolders` URL still applies.

#### Changes

* Added a shared `PanoramaPublicSchema.limitContainerScope` helper that restricts any container filter wider than the current folder and subfolders down to CurrentAndSubfolders.
* `PanoramaPublicSchema.createView` calls the helper for the `IsotopeModifications` and `StructuralModifications` grids.
* `ModificationsView.getContainerFilter` calls the same helper, closing the same gap on the standalone web parts. It is a no-op on the experiment details page, which already pins the scope.

#### Tests

* `PanoramaPublicModificationsTest.verifyModificationQueryScopeIsRestricted` checks that the Folder Filter menu does not include All Folders, and that a query.containerFilterName=AllFolders URL is restricted to CurrentAndSubfolders.
* `PanoramaPublicModificationsTest.verifyModificationWebPartScopeIsRestricted` drives the standalone web part URL and checks the same restriction.

### DIFF
--- a/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicSchema.java
+++ b/panoramapublic/src/org/labkey/panoramapublic/PanoramaPublicSchema.java
@@ -92,6 +92,11 @@ public class PanoramaPublicSchema extends UserSchema
     public static final String TABLE_EXPT_ISOTOPE_MOD_INFO = "ExperimentIsotopeModInfo";
     public static final String TABLE_ISOTOPE_UNIMOD_INFO = "IsotopeUnimodInfo";
 
+    // Queries behind the Structural and Isotope modification web parts. Both aggregate over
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
This is primarily to control bots, correct? If so, this should provide good protection. It would still be possible to create a custom SQL query that set a slow container filter, for example.

vagisha:
> This is primarily to control bots, correct? If so, this should provide good protection. It would still be possible to create a custom SQL query that set a slow container filter, for example.

Yes, this is to control bot traffic.

PR #667, opened 2026-07-26
Address security issues in the signup module

vagisha:
## Rationale
This PR addresses the security issues found in the signup module following an internal review. It tightens validation on self-service group changes, makes signup responses consistent, improves auditing of administrative changes, and avoids surfacing internal mail errors to callers.

#### Security review findings (signup module)

The review flagged six items in the signup module. Four are addressed by this PR; the two rate-limiting items were already handled by the kaptcha work in #638.

- **[Low-Med]** Self-service group change did not re-validate the resolved groups before changing membership, so a misconfigured transition rule could move a user into a more-privileged, site, or other-project group. *Addressed here.*
- **[Low]** Signup responses differed depending on whether the submitted address already had an account. *Addressed here.*
- **[Low]** Administrative configuration actions wrote no audit event on success (group-change rules and per-folder target group). *Addressed here.*
- **[Low]** The unauthenticated signup endpoint could send mail without rate limiting. *Addressed by kaptcha in #638.*
- **[Informational]** No rate limiting on the unauthenticated signup email endpoints. *Addressed by kaptcha in #638.*
- **[Informational]** Mail-send errors were returned to the caller. *Addressed here.*

## Related Pull Requests
 - #638.

## Changes
- **Group-change validation.** The self-service group change re-validates the resolved groups before changing any membership. Both must be project groups in the same project, and the target must grant no more than read access, with no write or administrative permission anywhere in the project tree. Requests that don't meet these constraints are refused with the generic NO_PERMISSIONS status (the specific reason is logged server-side), and successful moves are audited.
- **Consistent signup responses.** The signup endpoints return the same response whether or not an address already has an account, including when email delivery fails. The owner of an existing address receives a notice pointing to the password-reset flow, and the existing account is never modified.
- **Auditing.** The administrative configuration actions now write an audit event on success.
- **Error handling.** Mail-send errors are logged server-side, and callers receive a generic message.
- **Test.** Added `SignUpGroupChangeSecurityTest`, covering the group-change validation and its auditing.

 **Deployment note.** One DB-stored wiki page on the skyline website needs updating after deployment. The signup form should branch on status == 'SUCCESS'. The Skyline-daily register form needs no change, because a rejected group change now returns NO_PERMISSIONS, which the page already handles.

### DIFF
--- a/signup/src/org/labkey/signup/SignUpController.java
+++ b/signup/src/org/labkey/signup/SignUpController.java
@@ -36,7 +36,6 @@ import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.audit.ClientApiAuditProvider;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
-import org.labkey.api.data.ContainerType;
 import org.labkey.api.data.CoreSchema;
 import org.labkey.api.data.DbScope;
 import org.labkey.api.data.PropertyManager;
@@ -78,6 +77,7 @@ import org.springframework.validation.Errors;
 import org.springframework.validation.ObjectError;
 import org.springframework.web.servlet.ModelAndView;
 
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -233,8 +233,8 @@ public class SignUpController extends SpringActionController
             m.save();
             AuditLogService.get().addEvent(getUser(),
                     new ClientApiAuditProvider.ClientApiAuditEvent(ContainerManager.getRoot(),
-                            "Signup group-change rule added: members of group " + addGroupChangeForm.getOldgroup()
-                                    + " may move to group " + addGroupChangeForm.getNewgroup() + "."));
+                            "Signup group-change rule added: members of group " + groupLabel(addGroupChangeForm.getOldgroup())
+                                    + " may move to group " + groupLabel(addGroupChangeForm.getNewgroup()) + "."));
             return true;
         }
 
@@ -263,6 +263,8 @@ public class SignUpController extends SpringActionController
                 return false;
             WritablePropertyMap m = PropertyManager.getWritableProperties(SignUpModule.SIGNUP_GROUP_TO_GROUP, true);
             String existingRules = m.get(String.valueOf(oldgroup));
+            if(existingRules == null) // no rules configured for this source group - nothing to remove
+                return false;
             ArrayList<String> rules = new ArrayList<>(Arrays.asList(existingRules.split(",")));
             if(!rules.contains(String.valueOf(newgroup)))
                 return false;
@@ -270,13 +272,13 @@ public class SignUpController extends SpringActionController
             String newProperties = StringUtils.join(rules, ',');
             m.put(String.valueOf(oldgroup), newProperties);
             if(rules.isEmpty() || (rules.size() == 1 && rules.contains("")))
-                m.remove(oldgroup);
+                m.remove(String.valueOf(oldgroup)); // keys are stored as String.valueOf(oldgroup)
 
             m.save();
             AuditLogService.get().addEvent(getUser(),
                     new ClientApiAuditProvider.ClientApiAuditEvent(ContainerManager.getRoot(),
-                            "Signup group-change rule removed: members of group " + oldgroup
-                                    + " may no longer move to group " + newgroup + "."));
+                            "Signup group-change rule removed: members of group " + groupLabel(oldgroup)
+                                    + " may no longer move to group " + groupLabel(newgroup) + "."));
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


signup/src/org/labkey/signup/SignUpController.java
```diff
@@ -249,16 +263,22 @@ public boolean handlePost(AddGroupChangeForm addGroupChangeForm, BindException e
                 return false;
             WritablePropertyMap m = PropertyManager.getWritableProperties(SignUpModule.SIGNUP_GROUP_TO_GROUP, true);
             String existingRules = m.get(String.valueOf(oldgroup));
+            if(existingRules == null) // no rules configured for this source group - nothing to remove
+                return false;
```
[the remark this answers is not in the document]
vagisha: The null guard was added only to prevent an NPE when the source group has no rules. The false return matches the branch just below it, where !rules.contains(newgroup) also returns false when the rule isn't present, so this keeps the two "nothing to remove" cases consistent rather than changing pre-existing behavior. The action is only reached from the site-admin **Remove** link on rules that currently exist, so the "nothing to remove" path is a rare edge such as a stale page or double-submit.

signup/test/src/org/labkey/test/tests/signup/SignUpGroupChangeSecurityTest.java
```diff
@@ -0,0 +1,355 @@
...
+import org.labkey.remoteapi.query.SelectRowsResponse;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.WebTestHelper;
+import org.labkey.test.categories.External;
+import org.labkey.test.categories.MacCossLabModules;
+import org.labkey.test.util.APITestHelper;
+import org.labkey.test.util.ApiPermissionsHelper;
+import org.labkey.test.util.LogMethod;
+import org.labkey.test.util.PermissionsHelper.MemberType;
+import org.labkey.test.util.PermissionsHelper.PrincipalType;
```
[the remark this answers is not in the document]
vagisha: Fixed in https://github.com/LabKey/MacCossLabModules/pull/667/commits/9dbfa81ca8e1c26af2392334ac6fffe15137ee02

signup/src/org/labkey/signup/SignUpController.java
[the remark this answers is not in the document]
labkey-jeckels: Returning false without adding to the errors collection may result in a blank 200/success page. Consider calling `errors.addErrors()` for these two `return false;` branches if you think it's likely enough to want to help the user.

signup/src/org/labkey/signup/SignUpController.java
```diff
@@ -728,10 +758,46 @@ private static List<String> errorsToMessages(Errors errors)
...
+            return "site groups are not valid for self-service changes";
+        if (!newgroup.getContainer().equals(oldgroup.getContainer()))
+            return "target group is in a different project than the source group";
+        Container project = ContainerManager.getForId(newgroup.getContainer());
+        if (project == null)
+            return "project no longer exists";
+        // Reject a target group with admin permission anywhere in the project.
+        for (Container c : ContainerManager.getAllChildren(project))
+        {
+            if (c.hasPermission(newgroup, AdminPermission.class))
```
labkey-jeckels: This protects against admin escalation. Is that enough, or do you also need to check for trying to gain Editor or similar?

signup/src/org/labkey/signup/SignUpController.java
```diff
@@ -861,17 +916,34 @@ public ApiResponse execute(AddGroupChangeForm addGroupChangeForm, BindException
...
+            String denyReason = validateGroupChangeTarget(oldgroup, newgroup);
+            if (denyReason != null)
+            {
+                _log.warn("Rejected self-service group change for user {} (group {} -> {}): {}",
+                        user.getEmail(), addGroupChangeForm.getOldgroup(), addGroupChangeForm.getNewgroup(), denyReason);
+                // A configured rule points at a target the self-service flow must never grant (privileged,
+                // site, or other-project). Report this with its own status so a refused misconfiguration is
+                // distinguishable from an ineligible caller (NO_PERMISSIONS above). The specific reason is
+                // logged server-side only, not returned to the caller.
+                response.put("status", "TARGET_NOT_ALLOWED");
```
labkey-jeckels: Reveals the distinction between groups that exist and don't. Probably acceptable if it gives a usability win on the client side.

signup/test/src/org/labkey/test/tests/signup/SignUpGroupChangeSecurityTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+    @After
+    public void removePlantedRules()
+    {
+        // The transition rule map is site-global, so remove the rules this test planted even when project
+        // cleanup is skipped (clean=false). Runs after the test method.
+        if (_plantedRules.isEmpty())
+            return;
+
+        ensureSignedInAsPrimaryTestUser();
```
labkey-jeckels, line 330: Normally we'd avoid something like this in an `@After` because it prevents collecting the right screenshots as artifacts (since it does browser navigation). However, I think it's OK here because this is all API-testing, so there's no browser state worth capturing, right?

labkey-jeckels (approved):

signup/src/org/labkey/signup/SignUpController.java
vagisha: On skyline.ms the target group has read-only access, so anything more than that should not be required. 
Fixed in https://github.com/LabKey/MacCossLabModules/pull/667/commits/6aaeeb3121b5ceaab5da31de755c2e99ad298b7d.  `validateGroupChangeTarget` now rejects a target that grants write (Insert/Update/Delete) or admin permissions anywhere in the project tree.

signup/src/org/labkey/signup/SignUpController.java
vagisha: There isn't a client-side win, since we would show the same message that is shown for NO_PERMISSIONS. I collapsed the rejection to NO_PERMISSIONS. Fixed in https://github.com/LabKey/MacCossLabModules/commit/6aaeeb3121b5ceaab5da31de755c2e99ad298b7d

signup/test/src/org/labkey/test/tests/signup/SignUpGroupChangeSecurityTest.java
vagisha: Right. This test is entirely API-driven so there is no browser state worth capturing on failure.

